PR #78, opened 2024-02-11
Add Windows MSVC support

### LINKED ISSUE #75: Fails to compile on Windows with u32 != i32
Compiling on Windows results in lots of type mismatch errors like this:

```
error[E0308]: mismatched types
   --> llama-cpp-2\src\model.rs:335:11
    |
335 |     BPE = llama_cpp_sys_2::LLAMA_VOCAB_TYPE_BPE,
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `i32`

error[E0308]: mismatched types
   --> llama-cpp-2\src\model.rs:337:11
    |
337 |     SPM = llama_cpp_sys_2::LLAMA_VOCAB_TYPE_SPM,
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `i32`

...

error[E0308]: mismatched types
   --> llama-cpp-2\src\grammar.rs:410:28
    |
410 |                 ']' => Ok((u32::from(']'), rest)),
    |                            ^^^^^^^^^^^^^^ expected `i32`, found `u32`

error[E0308]: mismatched types
   --> llama-cpp-2\src\grammar.rs:414:17
    |
414 |             Ok((u32::from(c), &rest[c.len_utf8()..]))
    |                 ^^^^^^^^^^^^ expected `i32`, found `u32`
```

The types generated for enums look like `pub type llama_vocab_type = ::std::os::raw::c_int;`, which maps to `i32`, while this library expects those values to be `u32`.

This is essentially an issue with `bindgen` generating the native underlying type for `enum`s, which on Windows is different: https://github.com/rust-lang/rust-bindgen/issues/1966

I worked around it locally by adding `as _` in a couple places, which works fine, but I'm unsure if there's not a better approach.

Systemcluster:
Closes #75

### DIFF
--- a/.github/workflows/llama-cpp-rs-check.yml
+++ b/.github/workflows/llama-cpp-rs-check.yml
@@ -68,3 +68,15 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
       - name: Build
         run: cargo build
+  windows:
+    name: Check that it builds on windows
+    runs-on: windows-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+        with:
+          submodules: recursive
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@stable
+      - name: Build
+        run: cargo build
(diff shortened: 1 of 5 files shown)


MarcusDunn (changes requested):
Thanks for the PR - good change.

instead of `as _` could you reference the generated type (i.e. `ggml_type`) or similar?

MarcusDunn (approved):

PR #93, opened 2024-02-21
Process user defined tokens

SilasMarvin:
Thank you for creating and maintaining this crate.

I was doing some testing with https://huggingface.co/stabilityai/stable-code-3b/tree/main and noticed that I wasn't getting any of the spaces / tabs when I was working with Python code. After some testing I noticed we were just skipping over user defined tokens, and for this model, spaces and tabs are specialized user defined tokens. This edit seemed to fix that problem. I am pretty new to Llama cpp so if there is a better way to do this, please let me know.

### DIFF
--- a/llama-cpp-2/src/model.rs
+++ b/llama-cpp-2/src/model.rs
@@ -210,7 +210,7 @@ impl LlamaModel {
         }
 
         match self.token_type(token) {
-            LlamaTokenType::Normal => {}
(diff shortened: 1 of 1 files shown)


MarcusDunn:
perfect - the token type stuff was mostly guesswork on my part so if this change fixes something on your end it's likely the way to go. Thanks for the contribution!

MarcusDunn (approved):

PR #96, opened 2024-02-22
Working build.rs for apple metal

SilasMarvin:
This build seems to be working for me on mac with an m2 chip. I haven't tested on linux or anything else. The bench fails to compile, but the simple example works and and the tests with my local code depending on the crate are working. There are probably some changes that need to be made for linux. I'm not really sure as bindgen and compilation in general is outside the scope of what I normally work with.

Note this is still a work in progress and I need to clean a few things up. I wouldn't merge it yet


MarcusDunn:
I'm going to merge in #8 and do some testing - if you want to do some further cleanup feel free, but if it works for me I'll merge it to main.

PR #112, opened 2024-02-27
Expose functions `llama_load_session_file` and `llama_save_session_file`

zh217:
These two functions are useful for persisting sessions and are currently not exposed in the Rust wrapper.


MarcusDunn (approved):

MarcusDunn:
thanks, looks good.

PR #116, opened 2024-02-28
Expose the complete API for dealing with KV cache and states

zh217:
This pull request exposes the complete public C API for dealing with KV cache and internal states in Rust. I need this for writing efficient custom decoder, and it might also be useful for others.

The added method `copy_kv_cache_seq` is almost the existing `copy_cache`, but the latter is missing one argument. So I have added a note in the documentation of the latter.

### DIFF
--- a/llama-cpp-2/src/context/kv_cache.rs
+++ b/llama-cpp-2/src/context/kv_cache.rs
@@ -6,8 +6,6 @@ use crate::context::LlamaContext;
 impl LlamaContext<'_> {
     /// Copy the cache from one sequence to another.
     ///
-    /// Equivalent to `copy_kv_cache_seq` with `p0` set to `None` and `p1` set to `Some(size)`.
-    ///
     /// # Parameters
     ///
     /// * `src` - The sequence id to copy the cache from.
--- a/llama-cpp-2/src/context/kv_cache.rs
+++ b/llama-cpp-2/src/context/kv_cache.rs
@@ -1,9 +1,12 @@
 //! utilities for working with the kv cache
 
+use std::num::NonZeroU8;
 use crate::context::LlamaContext;
 
 impl LlamaContext<'_> {
     /// Copy the cache from one sequence to another.
+    ///
+    /// Equivalent to `copy_kv_cache_seq` with `p0` set to `None` and `p1` set to `Some(size)`.
     ///
     /// # Parameters
     ///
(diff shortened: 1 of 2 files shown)


llama-cpp-2/src/context/kv_cache.rs
```diff
@@ -31,4 +54,181 @@ impl LlamaContext<'_> {
...
+            )
+        }
+    }
+
+    /// Returns the largest position present in the KV cache for the specified sequence
+    ///
+    /// # Parameters
+    ///
+    /// * `seq_id` - The sequence id to get the max position for
+    pub fn kv_cache_seq_pos_max(&self, seq_id: i32) -> i32 {
```
MarcusDunn, line 128: What does a negative seq_id mean here (and elsewhere) Is it ever valid / meaningful?

llama-cpp-2/src/context/kv_cache.rs
```diff
@@ -31,4 +54,181 @@ impl LlamaContext<'_> {
...
+    /// * `p1` - The end position of the cache to update. If `None`, the entire cache is updated starting from [p0].
+    /// * `d` - The factor to divide the positions by
+    pub fn kv_cache_seq_div(&mut self, seq_id: i32, p0: Option<u16>, p1: Option<u16>, d: NonZeroU8) {
+        unsafe {
+            llama_cpp_sys_2::llama_kv_cache_seq_div(
+                self.context.as_ptr(),
+                seq_id,
+                p0.map_or(-1, i32::from),
+                p1.map_or(-1, i32::from),
+                d.get() as i32,
```
MarcusDunn, line 118: Avoid `as`, prefer `i32:from` as it'll throw an error if `d`s type ever changes to something that could fail.

llama-cpp-2/src/context/session.rs
```diff
@@ -130,4 +130,29 @@ impl LlamaContext<'_> {
...
+
+    /// Set the state reading from the specified address
+    /// Returns the number of bytes read
+    pub unsafe fn set_state_data(&mut self, src: &[u8]) -> usize {
+        unsafe {
+            // we don't really need a mutable pointer for `src` -- this is a llama-cpp lapse,
+            // so we cast away the constness
+            llama_cpp_sys_2::llama_set_state_data(self.context.as_ptr(), src.as_ptr() as *mut u8)
+        }
+    }
```
MarcusDunn, line 157: make `src` mutable. I'll open a PR to llama.cpp to make it `const` on their end

MarcusDunn: you can document that `src` will be changed to be immutable pending upstream changes, I'll fix once PR is merged.

MarcusDunn (commented):
these are all great to add.

- remove or deprecate the old API
- document what negative numbers mean or change them to be unsigned (or if they are valid leave as is)
- nit: remove all `as` in favor of `TryFrom` or `From` (panicking is fine for now)

llama-cpp-2/src/context/session.rs
MarcusDunn: or wait until PR is merged and the nightly update llama.cpp goes though.

llama-cpp-2/src/context/kv_cache.rs
zh217: As far as I understand, `seq_id` is just a label, ultimately used in a `std::set`, so yes negative numbers are fine.

In the C++ code almost all such parameters actually refers to `typedef int32_t llama_seq_id`, however the existing Rust wrapper does not introduce a newtype (notably `LlamaBatch::add`).

Also, I don't think there is enough checking in the C++ code to check what is passed in makes sense (a proper check would be highly non-trivial and slow). Worse, passing in the wrong thing is not guaranteed to trigger anything noticable. So I'm not sure what is the best way to do in the Rust part.

MarcusDunn (approved):
Looks good. Pending on llama.cpp changes before I cut a release but thanks for all the work!

PR #118, opened 2024-02-29
add with_main_gpu to LlamaModelParams

danbev:
This commit adds a `with_main_gpu` method to `LlamaModelParams` which allows the main GPU to be set.


MarcusDunn:
Could you also add a `main_gpu` to allow it to be read as well?

MarcusDunn (approved):

PR #120, opened 2024-02-29
override model values

MarcusDunn:
This could use a second set of eyes, first time doing anything non-future based with `std::pin`, but I think it's mostly done correctly?


MarcusDunn:
working, api sucks tho. will take a second stab at it.

MarcusDunn:
This seems to be usable. If you don't need `kv_overrides` you don't have to deal with `pin`, no more generics. 

Currently forces a user to allocate the params on the heap if they want to mess with overrides - which kinda sucks (I do not know a good way to be generic over the type of pinned pointer, which *i think* is what I want).

PR #121, opened 2024-02-29
Add CPU Feature Support

Hirtol:
This PR will ensure SIMD implementations will be used by GGML when the _target_ supports them.

Mostly taken from the `llm` crate. If you want to verify whether it worked you can simply use `llama_cpp_sys_2::llama_print_system_info()`

Note that you'll need to specify a target which has the features available, e.g. `-Ctarget-cpu=ivybridge` to enable AVX only. The base x86_64 target in Rustc has _no_ SIMD features available at the moment.

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -68,7 +68,7 @@ fn main() {
         let compiler = build.get_compiler();
 
         if cfg!(target_arch = "i686") || cfg!(target_arch = "x86_64") {
-            let features = x86::Features::get();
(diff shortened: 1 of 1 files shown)


MarcusDunn (approved):
looks good!

Hirtol:
Turns out the `llm` build.rs was not entirely correct. After a little bit of research it turns out that neon should be enabled by default on 64 bit ARM targets. After compiling the crate on my PI4 I can confirm that it automatically gets enabled in such cases. Removing the `-mfpu` flag should fix the workflow failure on ARM.

PR #133, opened 2024-03-05
Add Embedding Related Functionality

Hirtol:
Recently `llama.cpp` has added the ability to generate embeddings using BERT related models. There have been some issues, namely differences between the embeddings generated in Python and the ones coming from `llama.cpp`, but with https://github.com/ggerganov/llama.cpp/pull/5796 being merged it seems to finally be stable.

This PR implements the functionality to allow a user to make use of these new features, it adds an example based loosely on the `llama.cpp` embeddings example. Most of the code was translated from either `llama.cpp` or `llama-cpp-python` (specifically, the `add_sequence` method in `LLamaBatch`).

A few additional changes snuck in as well, replacing a `println!` call with a `tracing` call to allow users to filter it out, alongside a fix for a failing doc-test.

### DIFF
--- a/llama-cpp-2/src/lib.rs
+++ b/llama-cpp-2/src/lib.rs
@@ -52,6 +52,8 @@ pub enum LLamaCppError {
     /// There was an error adding a token to a batch.
     #[error["{0}"]]
     BatchAddError(#[from] BatchAddError),
+    #[error(transparent)]
+    EmbeddingError(#[from] EmbeddingsError),
 }
 
 /// Failed to Load context
@@ -74,6 +76,13 @@ pub enum DecodeError {
     /// An unknown error occurred.
     #[error("Decode Error {0}: unknown")]
     Unknown(c_int),
+}
+
+/// When embedding related functions fail
+#[derive(Debug, Eq, PartialEq, thiserror::Error)]
+pub enum EmbeddingsError {
+    #[error("Embeddings weren't enabled in the context options")]
+    NotEnabled,
 }
 
 /// Decode a error from llama.cpp into a [`DecodeError`].
--- a/llama-cpp-2/src/lib.rs
+++ b/llama-cpp-2/src/lib.rs
@@ -83,6 +83,10 @@ pub enum DecodeError {
 pub enum EmbeddingsError {
     #[error("Embeddings weren't enabled in the context options")]
     NotEnabled,
+    #[error("Logits were not enabled for the given token")]
+    LogitsNotEnabled,
+    #[error("Can't use sequence embeddings with a model supporting only LLAMA_POOLING_TYPE_NONE")]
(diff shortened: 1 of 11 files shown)


llama-cpp-2/src/lib.rs
```diff
@@ -52,6 +52,8 @@ pub enum LLamaCppError {
     /// There was an error adding a token to a batch.
     #[error["{0}"]]
     BatchAddError(#[from] BatchAddError),
+    #[error(transparent)]
```
MarcusDunn, line 55: til!

llama-cpp-2/src/llama_backend.rs
```diff
@@ -68,6 +69,19 @@ impl LlamaBackend {
...
+        unsafe extern "C" fn void_log(
+            _level: ggml_log_level,
+            _text: *const ::std::os::raw::c_char,
+            _user_data: *mut ::std::os::raw::c_void,
+        ) {}
+
+        unsafe {
+            llama_cpp_sys_2::llama_log_set(Some(void_log), std::ptr::null_mut())
+        }
+    }
```
MarcusDunn, line 84: very nice change.

MarcusDunn (commented):
this looks good. one question about consolidating some unsafe.

llama-cpp-2/src/llama_batch.rs
```diff
@@ -90,7 +90,50 @@ impl LlamaBatch {
...
+                let seq_id_ptr = *self.llama_batch.seq_id.add(j);
+                seq_id_ptr.write(seq_id);
+                self.llama_batch.n_seq_id.add(j).write(1);
+
+                let write_logits = logits_all || i == n_tokens - 1;
+                self.llama_batch.logits.add(j).write(write_logits as i8);
+                if write_logits {
+                    self.initialized_logits.push(j as i32);
+                }
+            }
```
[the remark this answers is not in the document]
Hirtol: Yeah, can do. Doing it this way was to avoid redundant checks as much as possible (and avoid the potentially expensive `Vec::retain` call in `LLamaBatch::add`), but it might be a preemptive optimisation.

Alternatively we could extract the unsafe parts into a private function and just share that call between `add` and `add_sequence`, if you'd prefer that?

Hirtol:
Hm, the doc-test that I 'fixed' fails here.  `llama_token_type` seems to get translated to `c_uint` here, but `c_int` for me locally.

MarcusDunn:
fix that and I'll add windows to the test CI so I don't break it again. (`as llama_token_type` where needed should work fine)

MarcusDunn (approved):

MarcusDunn:
thanks for the quick turnaround. Just cut a new release.

PR #204, opened 2024-03-20
Added passthrough metadata information to help downstream projects build native libraries using llama.cpp build artifacts

anagri:
Added passthrough metadata information to help downstream projects build native libraries using llama.cpp build artifacts

Added INCLUDE pointing to llama.cpp source folder for header lookup 
Added OUT_DIR pointing to build output directory for object file lookup

This allows downstream projects access to llama.cpp header files, and llama.cpp object files to link with their own C-source and compile.

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -199,6 +199,9 @@ fn main() {
     bindings
         .write_to_file(out_path.join("bindings.rs"))
         .expect("failed to write bindings to file");
+    let llama_cpp_dir = PathBuf::from("llama.cpp").canonicalize().unwrap();
+    println!("cargo:INCLUDE={}", llama_cpp_dir.to_str().unwrap());
+    println!("cargo:OUT_DIR={}", out_path.to_str().unwrap());
 }
 
 // courtesy of https://github.com/rustformers/llm


MarcusDunn:
This looks good - is there somewhere you could point me to where `INCLUDE` and `OUT_DIR` are used (other projects, etc)?

https://doc.rust-lang.org/cargo/reference/build-scripts.html does not seem to mention it explicitly.

MarcusDunn:
(don't worry about checks - I broke those myself)

MarcusDunn:
https://kornel.ski/rust-sys-crate mentions it and it seems its pretty ad-hoc. Thanks for the PR!

anagri:
> This looks good - is there somewhere you could point me to where `cargo:INCLUDE` and `cargo:OUT_DIR` are used (other projects, etc)?
> 
> https://doc.rust-lang.org/cargo/reference/build-scripts.html does not seem to mention it explicitly.

I am new to rust as well. The above were suggested by ChatGPT when i was stuck extending llama.cpp with my own C-code. 

Glad you found the link - https://kornel.ski/rust-sys-crate where include is mentioned.

Thanks for the merge.

MarcusDunn:
I may end up removing the OUT_DIR one, I imagine using the environment variable is preferred.

PR #207, opened 2024-03-20
Fix failing tests on MacOS

anagri:
1. Fixing typos in build.rs for framework Accelerate.

2. Fixing doctest expectation based on llama.cpp implementation.
from llama.cpp:12847
```
#ifdef GGML_USE_METAL
    // note: we usually have plenty of VRAM, so by default offload all layers to the GPU
    result.n_gpu_layers = 999;
#endif
```
accordingly have doctest expectation based on target_os -
```
/// #[cfg(not(target_os = "macos"))]
/// assert_eq!(params.n_gpu_layers(), 0, "n_gpu_layers should be 0");
/// #[cfg(target_os = "macos")]
/// assert_eq!(params.n_gpu_layers(), 999, "n_gpu_layers should be 999");
```


MarcusDunn (approved):

MarcusDunn:
Thanks!

PR #212, opened 2024-03-20
Added a couple of additional CLI arguments for benchmarking purposes.

jasonmccampbell:
If you are interested, I added a couple of additional command line arguments to mirror the llama.cpp default binary. These help with automating benchmarking tests.

Specifically:
* Note that the prompt now must be preceded by -p or --prompt to allow the prompt to be read from a file (-f / --file)
* Added specification for the number of threads to use
* The seed and context size can be specified as well.

Draft for now. I'll check to see if there are any docs impact if these changes are of interest.

### DIFF
--- a/simple/src/main.rs
+++ b/simple/src/main.rs
@@ -32,8 +32,11 @@ struct Args {
     #[command(subcommand)]
     model: Model,
     /// The prompt
-    #[clap(default_value = "Hello my name is")]
-    prompt: String,
+    #[clap(short = 'p', long)]
+    prompt: Option<String>,
+    /// Read the prompt from a file
+    #[clap(short = 'f', long, help = "prompt file to start generation")]
+    file: Option<String>,
     /// set the length of the prompt + output in tokens
     #[arg(long, default_value_t = 32)]
     n_len: i32,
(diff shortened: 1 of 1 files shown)


MarcusDunn:
Totally of interest, no doc changes required - I basically just use simple as a smoke test to see if I broke something catastrophically.

jasonmccampbell:
Thanks, @MarcusDunn . Let me know if there's anything you'd like changed. And thank you for this project, hopefully I'll have an opportunity to contribute more in the future.

MarcusDunn:
I'll merge as is. Nitpick wise the non-prompt args (as those are mutually exclusive) should be defaulted via clap. This should result in easier to maintain help messages + less logic in main.

jasonmccampbell:
No rush on my end. I agree that clap defaults would make the code cleaner. I left them off in this case to avoid overriding (or duplicating) the core llama.cpp defaults and keep the behavior consistent. For example, llama.cpp defaults the number of threads to the number of cores in the system if unspecified. If you prefer the defaults are specified in clap, I'll switch it over.

MarcusDunn:
That's absolutely better, you're right.

MarcusDunn (approved):

PR #214, opened 2024-03-21
llava binding 

tinglou:
1. llava multimodel api 
2. ollama local repo access


tinglou:
Related PR
https://github.com/ggerganov/llama.cpp/pull/6079

MarcusDunn:
Thanks for the PR, it's well written and I appreciate the tests!

Unfortunately, I do not feel like I can maintain bindings to `examples` or `common` (you'll notice there is a rust version of grammar parsing for example).

If we can replace these with Rust, or if llama.cpp moves the required API's to `llama.h` (as they plan to do for `common/sampling.cpp`) I would reconsider llava support.

Issue #219, opened 2024-03-25
Do we have support for Vulkan ?


oddpxl:
I'm asking as I've just compiled my first Rust app for Android...

..and now curious to understand what it would take to compile it with llama-cpp-rs support for Android Vulkan :)
 
https://www.reddit.com/r/LocalLLaMA/comments/1adbzx8/as_of_about_4_minutes_ago_llamacpp_has_been/

MarcusDunn:
No vulkan support at the moment.

I think the best bet is going to be using cmake to build `llama.cpp` - unfortunately adding an extra build time dep, but it simplfies maintenance a fair bit for all the platforms. See [#58](https://github.com/utilityai/llama-cpp-rs/issues/211) for an idea of what that would look like.

I will not be putting any effect this direction, all we use internally is linux + cuda so justifying work elsewhere is a tough sell. 

Contributions for vulkan are welcome though.

oddpxl:
Does this mean llama-cpp-rs won't have Metal support going forward ?

llama-cpp-rs seems to be the fastest Rust LLM out there at the moment...Candle isn't even close :)

No worries I completely understand since your daily need is Cuda and not Metal / Vulkan / Other etc

MarcusDunn:
https://github.com/utilityai/llama-cpp-rs/pull/202 I believe broke metal support which was in 0.1.42. (0.1.41 should be fine)

I expect metal support to be working once agin upon replacing `cc` with `cmake` in the `build.rs`, I'll take a stab at it sometime this week I think, but no promises :)

MarcusDunn:
feel free to give #221 a try. No clue what the defaults for mac are (but a cursory glance makes me think it might *just work™)

oddpxl:
I tried #221 and it first asked for libllama.dylib ( temp fix with some ructc compiler directive to set rpath ) ...then for ggml-metal.metal ( fixed that by copying it to the same folder ) ...then error: 'ggml-common.h' not found - back to square one :)

Unfortunately 0.1.41 don't compile either because of an old bug:

error[E0609]: no field `n_max_seq` on type `llama_kv_cache_view`
   --> /Users/odd/.cargo/registry/src/index.crates.io-6f17d22bba15001f/llama-cpp-2-0.1.41/src/context/kv_cache.rs:241:63

 usize::try_from(self.view.n_cells * self.view.n_max_seq)
                                                                          ^^^^^^^^^ unknown field
    
[...]

PR #221, opened 2024-03-26
use cmake instead of `cc`

MarcusDunn:
bear bones implementation of swapping over to `cmake`.

I ran into some dastardly issues with doctests not linking shared libraries, it been [fixed](https://github.com/rust-lang/cargo/pull/13490) in the latest beta so I will not merge this until that's stable (2 May, 2024)

~Time permitting I will work on testing Metal.~ works

# TODO:

- [ ] Static Linking https://github.com/utilityai/llama-cpp-rs/pull/221#discussion_r1545450457
- [ ] Set Targets https://github.com/utilityai/llama-cpp-rs/pull/221#discussion_r1545451594


MarcusDunn:
seems metal_hack is still needed

MarcusDunn:
metal hack complete

MarcusDunn:
this should now fix #219

llama-cpp-sys-2/build.rs
```diff
@@ -1,187 +1,41 @@
...
+            } else {
+                "OFF"
+            },
+        )
+        .define("LLAMA_VULKAN", if cfg!(feature = "vulkan") {
+            "ON"
+        } else {
+            "OFF"
+        })
+        .define("BUILD_SHARED_LIBS", "ON")
```
Hirtol, line 30: It would be appreciated to have a `static` feature for those who prefer static linking.

llama-cpp-sys-2/build.rs
```diff
@@ -1,187 +1,41 @@
...
-
-        ggml.define("GGML_USE_CUBLAS", None);
-        ggml_cuda.define("GGML_USE_CUBLAS", None);
-        llama_cpp.define("GGML_USE_CUBLAS", None);
-    }
-
-    for build in [&mut ggml, &mut llama_cpp] {
-        let compiler = build.get_compiler();
-
-        if cfg!(target_arch = "i686") || cfg!(target_arch = "x86_64") {
```
Hirtol, line 74: Note that by default the CMake build will assume `target-cpu=native` (`LLAMA_NATIVE`), so you'll lose some portability by removing this. I don't mind making a PR to adapt this feature detection to the CMake build after this is merged, if you'd prefer.

MarcusDunn: I can re-add the feature detection (doesn't seem fair to delete your code and expect you to add it back!)

How does the feature detection as we had it differ from targeting native as exists in the cmake file?

What sort of portability do we lose out on?

Hirtol: The previous feature detection worked based on the target's features, e.g., if you want to ensure almost any modern CPU can run your software you can use `-Ctarget-cpu=ivybridge` as a compiler flag to enable _just_ `avx` and `sse3`, which any CPU made post 2012 should support. 

With the `LLAMA_NATIVE` flag it depends on the CPU of the one compiling the software, so it will enable all features and ignore your `-Ctarget-cpu=ivybridge` setting. Or in turn, it might not enable AVX2 if it was compiled on an older CPU.

To replicate the old behaviour it's just a case of passing `LLAMA_AVX/2=ON/OFF` etc, to the cmake build instead of the compiler specific flags (and turn off `LLAMA_NATIVE`). Similar to what `llama.cpp` does in their CI.

MarcusDunn: Thanks, that seems worth the effort. You're welcome to make a PR to this branch (won't be merged for a while). 

Otherwise I'll make sure it's done before this gets merged into main.

PR #224, opened 2024-03-27
MacOS currently fails on: #include "ggml-common.h"

derrickpersson:
Screenshot 2024-03-27 at 12 48 29 PM

- This adds onto the current hack of inserting the file as a string by also inserting the header dependency of the `metal` file into that string as well.

- Tried other approaches like including the `ggml-common.h` file, but ran into obtuse linker errors.


MarcusDunn:
Horrifying, thanks for the fix!

MarcusDunn:
no worries about windows CI, it makes Mac look well supported :)

oddpxl:
Release 0.1.45 compiles perfectly on my Mac M1 max - delivers 115 t/s using Gemma 2B

Many thanks @derrickpersson & @MarcusDunn  !

PR #232, opened 2024-04-01
fix: multi-byte utf8 decoding error

jiabochao:
The `llama_token_to_piece()` function returns bytes that may not always represent a complete UTF-8 sequence. 

- https://github.com/ggerganov/llama.cpp/discussions/3114

When encountering multi-byte UTF-8 characters, there's a possibility that a character could be split and returned in two parts. 

https://github.com/utilityai/llama-cpp-rs/blob/48cc14a2fc6c63e958ea0b966907983e71822232/llama-cpp-2/src/model.rs#L238

This would result in an error when attempting to convert these bytes into a string using `String::from_utf8`. 

```
        #ggml_metal_free: deallocating
Error: FromUtf8Error incomplete utf-8 byte sequence from index 1

Caused by:
    incomplete utf-8 byte sequence from index 1
```

My solution is to introduce the `model.token_to_bytes` function that directly returns bytes, and then to apply `encoding_rs` for decoding throughout the entire loop.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -282,6 +282,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a357d28ed41a50f9c765dbfe56cbc04a64e53e5fc58ba79fbc34c10ef3df831f"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


jiabochao:
The error can be reproduced using the following command:

```shell
cargo run --release --bin simple -- -p "Python如何实现冒泡排序:" --n-len 2048 local "qwen1_5-7b-chat-q2_k.gguf"
```

MarcusDunn (approved):
Lgtm

MarcusDunn:
this should be available in 0.1.46

PR #246, opened 2024-04-10
new feature `metal` to turn on/off metal framework on macos

tinglou:
1. Metal framework doesn't work on MacOS X86_64, a new feature 'metal' is added.
If the target is 'macos' and 'aarch64', the feature is on by default.

2. `llama.cpp` and `ollama` have the same issue, only  `Acelerate`(pure CPU) works on macos.

### DIFF
--- a/llama-cpp-2/Cargo.toml
+++ b/llama-cpp-2/Cargo.toml
@@ -15,7 +15,11 @@ tracing = { workspace = true }
 
 [features]
 cublas = ["llama-cpp-sys-2/cublas"]
+metal = ["llama-cpp-sys-2/metal"]
 sampler = []
+
+[target.'cfg(all(target_os = "macos", any(target_arch = "aarch64", target_arch = "arm64")))'.dependencies]  
+llama-cpp-sys-2 = { path = "../llama-cpp-sys-2", features=["metal"], version = "0.1.48" }
 
 [lints]
 workspace = true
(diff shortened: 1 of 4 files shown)


llama-cpp-2/Cargo.toml
```diff
@@ -15,8 +15,12 @@ tracing = { workspace = true }
 
 [features]
 cublas = ["llama-cpp-sys-2/cublas"]
+metal = ["llama-cpp-sys-2/metal"]
 sampler = []
 
+[target.'cfg(all(target_os = "macos", any(target_arch = "aarch64", target_arch = "arm64")))'.dependencies]  
+llama-cpp-sys-2 = { path = "../llama-cpp-sys-2", features=["metal"], version = "0.1.48" }
```
MarcusDunn, line 22: what does this do?

simple/Cargo.toml
```diff
@@ -14,6 +14,7 @@ encoding_rs = { workspace = true }
 
 [features]
 cublas = ["llama-cpp-2/cublas"]
+metal =  ["llama-cpp-2/metal"]
```
MarcusDunn, line 17: I think this should be default.

llama-cpp-2/Cargo.toml
tinglou: if the target is `macos` and `arm64`, feature `metal` is on, otherwise is off.

that is, macos user of crate `llama-cpp-2` needn't care about the arch (apple or intel) .

simple/Cargo.toml
tinglou: the feature `metal` is hidded by `llama-cpp-2`, and `simple` could turn on/off it manually.

MarcusDunn:
thanks for the clarification - should be available in next release.

PR #323, opened 2024-05-31
Updated llama-cpp (bot)

MarcusDunn:


SilasMarvin:
Can we merge this in? We are currently behind llama.cpp by a bit

PR #326, opened 2024-06-01
Switched build script

SilasMarvin:
I found this build script: https://github.com/edgenai/llama_cpp-rs/blob/main/crates/llama_cpp_sys/build.rs Which has support for more architectures and features.

I swapped out the current build script with this one. I tested and it works on Apple and fixes a metal not found warning. I haven't tested it on Linux with Cuda yet.

What do you think about swapping out the build script with this new one?

### DIFF
--- a/llama-cpp-sys-2/src/lib.rs
+++ b/llama-cpp-sys-2/src/lib.rs
@@ -4,27 +4,14 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 
-use std::fmt::{Debug, Formatter};
+// [`ash`] is only included to link to the Vulkan SDK.
+#[allow(unused)]
+#[cfg(feature = "vulkan")]
+use ash;
+
+// [`cudarc`] is only included to link to CUDA.
+#[allow(unused)]
+#[cfg(feature = "cuda")]
+use cudarc;
+
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
-
-impl Debug for llama_grammar_element {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        fn type_to_str(r#type: llama_gretype) -> &'static str {
-            match r#type {
-                LLAMA_GRETYPE_END => "END",
-                LLAMA_GRETYPE_ALT => "ALT",
-                LLAMA_GRETYPE_RULE_REF => "RULE_REF",
-                LLAMA_GRETYPE_CHAR => "CHAR",
-                LLAMA_GRETYPE_CHAR_NOT => "CHAR_NOT",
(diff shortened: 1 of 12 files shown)


MarcusDunn:
I'll have to benchmark with CUDA first, but if everything checks out I'll merge.

SilasMarvin:
> I'll have to benchmark with CUDA first, but if everything checks out I'll merge.

Let me know how it goes and if there are alterations we need to make

SilasMarvin:
Anything I can do to help with this? I do have a package with this as a dependency and am waiting on this new build script before pushing it out.

MarcusDunn:
I'm back in the office tomorrow - I can get it merged and a release cut then.

llama-cpp-sys-2/src/lib.rs
MarcusDunn, line 1: not sure what this change is doing.

llama-cpp-sys-2/src/lib.rs
```diff
@@ -4,27 +4,14 @@
...
+#[cfg(feature = "cuda")]
+use cudarc;
 
-        f.debug_struct("llama_grammar_element")
-            .field("type", &type_to_str(self.type_))
-            .field("value", &self.value)
-            .finish()
-    }
-}
+include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
```
MarcusDunn, line 17: I'm not certain these changes are required and / or doing anything?

MarcusDunn:
cuda does not work for the moment. I think it is because there's a feature name mismatch. We should rename the `cublas` feature to `cuda` (llama.cpp) did this - we should follow them (and that's what the build script uses)

SilasMarvin:
> cuda does not work for the moment. I think it is because there's a feature name mismatch. We should rename the `cublas` feature to `cuda`. llama.cpp did this - we should follow them (and that's what the build script uses)

Did my best to search and replace here: https://github.com/utilityai/llama-cpp-rs/pull/326/commits/6b0f923b2a0ba5befb07ee6b5fd337f967663b5f

MarcusDunn:
I'm now getting:
```
error[E0432]: unresolved import `cudarc`
```
I'd prefer to not add it as a dependency especially to the `sys` crate. Could you remove it and add the linking logic to the build script (you can copy from the old build script)

Also remove the `use ash` - it will result in the same error with vulkan enabled.

SilasMarvin:
> Update build script to remove cudarc

Updated to what I think you wanted. It doesn't build for me, but the old one doesn't build for me either. https://github.com/utilityai/llama-cpp-rs/pull/326/commits/09c7b612cf2945e47f27c6f57b6f005eac49000e

MarcusDunn:
106.13t/s!

PR #337, opened 2024-06-10
Add hipBLAS feature and fix build script

AdamNiederer:
Attempts to copy the CMake build steps as much as possible. GPU is detected when running but on my setup appears to segfault when loading the model. I'm at a bit of a loss on the segfault but this is definitely better than before.


MarcusDunn (approved):

MarcusDunn:
> Attempts to copy the CMake build steps as much as possible. GPU is detected and models are loaded into VRAM but it appears to segfault when doing anything else. I'm at a bit of a loss on the segfault but this is definitely better than before.

So to be clear, if I merge this, it still segfaults with HIP, just later?

AdamNiederer:
> So to be clear, if I merge this, it still segfaults with HIP, just later?

Correct, previously it would not build at all; now it builds, links, and can run some things, but will quickly segfault on the GPU when attempting to do anything of consequence. I've been unable to pinpoint why so hopefully somebody more knowledgeable can use this as a head start.

MarcusDunn:
Do you befit from having this merged, or is it fine here as a starting point for someone who wants to bring it across the finish line?

I'd prefer to not give the impression of HIP support if it's known to not work, but if it's useful then I don't mind merging.

AdamNiederer:
I'm fine with leaving keeping a local dependency for now since I'm not able to get my workload running. I'll mark this as a draft

PR #351, opened 2024-06-18
make LlamaTokenAttr a bitflags

estokes:
As discussed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,56 +10,57 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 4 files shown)


MarcusDunn (approved):

estokes:
On windows for some reason bindgen decides to use a c_int instead of a c_uint. I checked that the constants are always positive and then added a cast to u32, so it now builds on both platforms.

marek-hradil:
@estokes @MarcusDunn Hi, could you provide more details on why the `if` branch is important in this case?
Im struggling a bit to understand why would you like to prevent this method (and then consequently `token_to_str`) from outputting how does the eos and bos token look like - this did in fact cause an issue in our codebase, which depends on `llama-cpp-rs` and I would like to remove it, but also not break stuff for others.

PR #370, opened 2024-06-29
Fix character range grammar parsing

tommyip:
GBNF grammar for character ranges is not parsed correctly. For example, `[a-z]` is currently parsed as `("a" | "-" | "z")`.

This PR updates it to match upstream:
```cpp
if (pos[0] == '-' && pos[1] != ']') {
    if (!pos[1]) {
        throw std::runtime_error("unexpected end of input");
    }
    auto endchar_pair = parse_char(pos + 1);
         pos          = endchar_pair.second;
    out_elements.push_back({LLAMA_GRETYPE_CHAR_RNG_UPPER, endchar_pair.first});
}
```
https://github.com/ggerganov/llama.cpp/blob/72272b83a3878e91251218c981b4c6ec16c33912/common/grammar-parser.cpp#L241-L248

### DIFF
--- a/llama-cpp-2/src/grammar.rs
+++ b/llama-cpp-2/src/grammar.rs
@@ -294,8 +294,8 @@ impl ParseState {
                         type_: gre_type,
                         value: c as _,
                     });
-                    if rest.starts_with("-]") {
-                        let (c, r) = Self::parse_char(rest)?;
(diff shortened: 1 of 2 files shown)


MarcusDunn:
Kudos to you for jumping into that side of the code. Looks good to me.

MarcusDunn (approved):

PR #365, opened 2024-06-25
Enable native and vulkan features

estokes:
I've had pretty good luck with vulkan on Intel arc integrated GPUs, it's faster and more power efficient than using avx2-vnni on the CPU. I'm not sure if it was an oversight that the feature wasn't enabled, or if it just had not been tested.

I've tested it on ubuntu 22.04 on an Intel GPU. I'll try it on windows as soon as I have time.

PR #373, opened 2024-06-29
Expose CPU extension features in llama-cpp-sys-2, and automatically enable them in llama-cpp-2

babichjacob:
Currently, `llama-cpp-sys-2`'s build script references cargo features for CPU instruction set extensions:
https://github.com/utilityai/llama-cpp-rs/blob/78de7a12e906cdd02bc00af0e29a0bc29d6b6f6a/llama-cpp-sys-2/build.rs#L262-L316

but they have not been declared in its `Cargo.toml`. This PR does that (which should reduce unnecessary warnings during compilation).

I have not included features for BLAS / hipBLAS / CLBlast, CUDA, or compat since they're not related to instruction set extensions (the focus of this PR) and they're not low hanging enough of fruit for me since I don't know how they work. I also have not included `native` or `vulkan` for the same reason, but at least those are handled in #365 (and sorry if that needs to be rebased from this).

Instead of exposing the features a second time as a pass-through in `llama-cpp-2` (nor a third time in the `simple` example / binary or in any dependent user's crate), `llama-cpp-2` will now use [target-specific dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies) to enable `llama-cpp-sys-2`'s newly exposed CPU extension features. I figured this arrangement made the most sense because, as far as I'm aware, it doesn't add additional build requirements (the way e.g. CUDA or Vulkan would) so there shouldn't be a chance of compilations that used to succeed before this PR beginning to fail with this change.

Users of `llama-cpp-2` can ensure their available instruction set extensions are taken advantage of by either [specifying  `-C target-cpu=native` in their `RUSTFLAGS`](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-cpu), or by adding `llama-cpp-sys-2` as a dependency and enabling its `features` as they wish (which is similar enough of an experience to if `llama-cpp-2` had exposed the features again that it made me choose not to do it that way).

(I tested seeing a speedup from (auto-)enabling AVX2 on my own computer).

Let me know if this is helpful, thanks!

### DIFF
--- a/llama-cpp-2/Cargo.toml
+++ b/llama-cpp-2/Cargo.toml
@@ -19,6 +19,19 @@ cuda = ["llama-cpp-sys-2/cuda"]
 metal = ["llama-cpp-sys-2/metal"]
 sampler = []
 
+[target.'cfg(target_feature = "avx")'.dependencies]
+llama-cpp-sys-2 = { path = "../llama-cpp-sys-2", features = ["avx"] }
+[target.'cfg(target_feature = "avx2")'.dependencies]
+llama-cpp-sys-2 = { path = "../llama-cpp-sys-2", features = ["avx2"] }
+[target.'cfg(target_feature = "avx512vbmi")'.dependencies]
+llama-cpp-sys-2 = { path = "../llama-cpp-sys-2", features = ["avx512_vmbi"] }
+[target.'cfg(target_feature = "avx512vnni")'.dependencies]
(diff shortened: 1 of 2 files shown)


MarcusDunn (approved):
LGTM! thanks for the PR.

MarcusDunn:
will run tests with #365 then I'll merge.

Issue #393, opened 2024-07-09
0.1.61 failed to publish to crates.io


babichjacob:
I saw the publish job failed https://github.com/utilityai/llama-cpp-rs/actions/runs/9863063828/job/27234999318 and I think it's failing because of the changes in my PR that don't specify the version: https://github.com/utilityai/llama-cpp-rs/pull/373/files#diff-d23211a10497dbdcafcf9aca8827b92ad02db23a2859687844cd0d50f776c794R24-R38

Sending a PR to fix (if my diagnosis is correct) soon..

PR #404, opened 2024-07-15
Simplify CPU features: just detect them in build.rs instead of exposing them as crate features

babichjacob:
Since #373, where I wrote
> Instead of exposing the features, ... I figured this arrangement made the most sense because, as far as I'm aware, it doesn't add additional build requirements (the way e.g. CUDA or Vulkan would) so there shouldn't be a chance of compilations that used to succeed before this PR beginning to fail with this change.
Users of llama-cpp-2 can ensure their available instruction set extensions are taken advantage of by either [specifying -C target-cpu=native in their RUSTFLAGS](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-cpu) ...

I had an epiphany today that it would remove unnecessary code to just detect the CPU features in `build.rs` instead of exposing them as crate features. (and it would've prevented #393 lol oops) Since this part of the build script wasn't actually working yet, it was probably the original author's intention to check CPU features rather than crate features anyway.

(I don't have the means to test AVX512).

### DIFF
--- a/llama-cpp-2/Cargo.toml
+++ b/llama-cpp-2/Cargo.toml
@@ -22,35 +22,6 @@ vulkan = ["llama-cpp-sys-2/vulkan"]
 native = ["llama-cpp-sys-2/native"]
 sampler = []
 
-[target.'cfg(target_feature = "avx")'.dependencies]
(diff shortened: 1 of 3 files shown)


MarcusDunn (approved):

MarcusDunn:
This is a considerable improvement, thanks!

PR #410, opened 2024-07-20
Enlarge the buffer in `LlamaModel::apply_chat_template`

### LINKED ISSUE #409: Llama 3 templates get too small of a buffer in `LlamaModel::apply_chat_template`
Reproduction in https://github.com/babichjacob/llama-cpp-rs/commit/710d0824e35d1bdc75533c1691153c3aa9549ffa
specifically https://github.com/babichjacob/llama-cpp-rs/blob/710d0824e35d1bdc75533c1691153c3aa9549ffa/bug-repro/src/main.rs#L35-L37

Sorry this reproduction has to be a binary requiring you to have a full-on Llama 3 model on your computer -- I'll write another issue for thoughts on reworking this part of `llama-cpp-2`.

So, the code shows loading `LlamaModel` from a Meta Llama 3 GGUF (or finetune that doesn't change the template, e.g. I got it from [Salesforce's `SFR-Iterative-DPO-LLaMA-3-8B-R-Q6_K.gguf`](https://huggingface.co/bartowski/LLaMA3-iterative-DPO-final-GGUF)) then trying to use the`apply_chat_template` method on some example chat history. This results in an `Err(ApplyChatTemplateError::BuffSizeError)`. (Unwrapping it then causes a panic.) The bug occurs regardless of whether the `add_ass` argument (for adding an assistant prompt) is true or false.

babichjacob:
Fixes #409

Per this comment I saw in the code,
https://github.com/utilityai/llama-cpp-rs/blob/64409ebaecb815e61a0c2747c51f858949c4d758/llama-cpp-2/src/model.rs#L485-L489

this PR just increases the buffer size from times 2 to times 4 (cause 3 wasn't enough to resolve the issue on my end).

### DIFF
--- a/llama-cpp-2/src/model.rs
+++ b/llama-cpp-2/src/model.rs
@@ -454,7 +454,7 @@ impl LlamaModel {
         let message_length = chat.iter().fold(0, |acc, c| {
             acc + c.role.to_bytes().len() + c.content.to_bytes().len()
         });
-        let mut buff: Vec<i8> = vec![0_i8; message_length * 2];
+        let mut buff: Vec<i8> = vec![0_i8; message_length * 4];
 
         // Build our llama_cpp_sys_2 chat messages
         let chat: Vec<llama_cpp_sys_2::llama_chat_message> = chat


MarcusDunn:
This is fine, ideally we'd have a method that allows the caller to set the buffer size. (And have the more common one call it with a sensible default)

PR #412, opened 2024-07-21
BREAKING: Add a variant to `ChatTemplateError` and `LlamaModel::get_chat_template` for when the user specifies too small of a buffer

### LINKED ISSUE #411: `LlamaModel::get_chat_template` panics if the buffer size passed is too small for the template to fit in
Reproduction in https://github.com/babichjacob/llama-cpp-rs/commit/35c1de9970e35256dee592391da879a99ad34113
specifically https://github.com/babichjacob/llama-cpp-rs/commit/35c1de9970e35256dee592391da879a99ad34113#diff-d7652fc11ef443b44fefcec061161ffc10fa5b19332c1d345cd3ad47465786e8R26-R35

So, the code shows loading `LlamaModel` (any model / GGUF will do) then trying to use the`get_chat_template` method with an insufficient buffer size (just 15 bytes long). This results in a panic
```
thread 'main' panicked at llama-cpp-2\src\model.rs:384:13:
assertion `left == right` failed: llama.cpp guarantees that the returned int 344 is the length of the string 14 but that was not the case
  left: 344
 right: 14
```
but it could be changed to be an `Err` so calling code can recover: there isn't a memory vulnerability since [`llama.cpp` won't write outside of the buffer](https://github.com/ggerganov/llama.cpp/blob/45f2c19cc57286eead7b232ce8028273a817aa4d/src/llama.cpp#L19520-L19529).

babichjacob:
Fixes #411 but is a BREAKING CHANGE because this `enum` is not marked as `#[non_exhaustive]`, so any `match`es on it will now fail to compile if they don't have a fallback arm, or will now behave differently if they do.

Now, instead of panicking, `get_chat_template` will return a `ChatTemplateError::BuffSizeError(usize)` where the `usize` is the smallest size of a buffer that would be sufficiently large enough for `llama.cpp` to be able to print the chat template into.

A user could call `get_chat_template` again with the suggested buffer size to then be able to get it without error. (The alternative being just hoping to pass a sufficiently large buffer even if wasteful). 

Feel free to close this pull request if there's a better solution. I expect to write a future issue about creating a wrapper around `llama_model_meta_val_str` that allocates so the user doesn't have to specify a buffer size anyway.

### DIFF
--- a/llama-cpp-2/src/lib.rs
+++ b/llama-cpp-2/src/lib.rs
@@ -64,6 +64,9 @@ pub enum LLamaCppError {
 /// There was an error while getting the chat template from a model.
 #[derive(Debug, Eq, PartialEq, thiserror::Error)]
 pub enum ChatTemplateError {
+    /// the buffer was too small.
+    #[error("The buffer was too small. However, a buffer size of {0} would be just large enough.")]
+    BuffSizeError(usize),
     /// gguf has no chat template
     #[error("the model has no meta val - returned code {0}")]
     MissingTemplate(i32),
--- a/llama-cpp-2/src/model.rs
+++ b/llama-cpp-2/src/model.rs
@@ -370,22 +370,28 @@ impl LlamaModel {
         let chat_ptr = chat_temp.into_raw();
         let chat_name = CString::new("tokenizer.chat_template").expect("no null bytes");
 
-        let chat_template: String = unsafe {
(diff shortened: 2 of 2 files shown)


MarcusDunn (approved):

MarcusDunn:
The wrapper would be a welcome addition. This is good.

PR #400, opened 2024-07-10
Update llama cpp [MAC TESTING WANTED]

MarcusDunn:

PR #417, opened 2024-07-26
macOS `build.rs` fixes to align w ggerganov/llama.cpp@f3f6542

brittlewis12:
Thanks for your work establishing and maintaining these bindings!

Re: https://github.com/utilityai/llama-cpp-rs/pull/400#issuecomment-2253166089, these changes align Metal compilation with pattern established in the CUDA compilation code.

I also noticed what I think was a small typo in an environment variable being read to enable llama header location overrides. Happy to exclude that if I was mistaken!

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -687,7 +687,7 @@ fn main() {
         println!("cargo:rustc-link-lib=llama");
         println!("cargo:rustc-link-lib=ggml");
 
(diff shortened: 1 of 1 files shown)


MarcusDunn:
thanks for the quick PR!

MarcusDunn (approved):

brittlewis12:
make a quick change to rebase relative to main, thanks for your quick feedback!

Issue #438, opened 2024-08-05
Support for encoder-decoder models (T5).


cyanic-selkie:
Hi, `llama.cpp` added support for T5, but this requires the `encode` method. Will this be added?

PR #439, opened 2024-08-05
add bindings for lora and encode

cyanic-selkie:
I made an issue at #438, however, since the changes were trivial I made this PR - it supports everything I need, but maybe it's a good idea to add bindings for some other related functionalities as well. Also, I added bindings for lora functionalities as well as the encode functionality.


MarcusDunn (approved):

MarcusDunn:
Thanks for the changes! Looks good to me

PR #466, opened 2024-08-20
README.md: Fix example call to the "simple" binary

Papierkorb:
Hello,

Just stumbled upon your library! The example call didn't work out of the box, the syntax is slightly different. So here's a small fix for this :)

Cheers!


MarcusDunn:
Thanks!

Issue #505, opened 2024-09-22
Proposal: Expose `llama_get_logits`-based method on Context in high level API


brittlewis12:
Today, llama-cpp-2 exposes `llama_get_logits_ith`-based `candidates_ith`.

In working with the underlying library, I have leaned on the slightly different  `llama_get_logits`. It seems that while the underlying library accepts `-1` as input to the `-ith` variant to accomplish the same as `llama_get_logits`, the current implementation of safely ensuring logits are initialized for the given `llama_pos` disallows `-1` as input as a side effect.

Honestly, the existing implementation is sound as is, and I'd prefer to use the other, slightly simpler seam either way.

I took a quick stab at what this could look like, based on the existing pattern of a lower-level `unsafe` call and a higher level wrapper that returns an unsorted `LlamaTokenData` iter for you.
- [Expose last-token candidates context method](https://github.com/brittlewis12/llama-cpp-rs/pull/1)

I [tweaked the existing](https://github.com/brittlewis12/llama-cpp-rs/compare/candidates_last...brittlewis12:llama-cpp-rs:candidates-last-simple?expand=1#diff-7319277c084bff5c625f04b23366571e88d4463276613007bee0d70b72cbc445) `simple` example to use this method without issue:
```
-            let candidates = ctx.candidates_ith(batch.n_tokens() - 1);
+            let candidates = ctx.candidates_last();
```

Thank you as ever for your maintenance efforts here @MarcusDunn!

---

a little more background [via llama.cpp](https://github.com/ggerganov/llama.cpp/blob/c35e586ea57221844442c65a1172498c54971cb0/include/llama.h#L844-L849):

> Token logits obtained from the last call to llama_decode()
> The logits for which llama_batch.logits[i] != 0 are stored contiguously in the order they have appeared in the batch.
> Rows: number of tokens for which llama_batch.logits[i] != 0
> Cols: n_vocab

MarcusDunn:
whoops, thought that was a PR, didn't mean to review your fork!

brittlewis12:
no problem 😆 will open it up properly on your end

brittlewis12:
re: https://github.com/brittlewis12/llama-cpp-rs/pull/1#pullrequestreview-2323689876
would removing “last” from the names, so `get_logits` & just `candidates`, be more in line with your preferred naming conventions?

MarcusDunn:
yeah, that matches what I've done elsewhere a more closely.

[...]

PR #510, opened 2024-09-25
Context and model enhancements

brittlewis12:
As mentioned in https://github.com/utilityai/llama-cpp-rs/issues/505#issuecomment-2372739691, various enhancements to Context & Model capablities:
- mirostat v1 sampling
- `llama_token_is_eog` equivalent
- `llama_kv_cache_seq_rm` Context method for kv cache manipulation
- `flash_attn`, `offload_kqv` context parameters

### DIFF
--- a/llama-cpp-2/src/context/kv_cache.rs
+++ b/llama-cpp-2/src/context/kv_cache.rs
@@ -2,7 +2,21 @@
 
 use crate::context::LlamaContext;
 use std::ffi::c_int;
-use std::num::NonZeroU8;
+use std::num::{NonZeroU8, TryFromIntError};
+
+/// Errors that can occur when attempting to prepare values for the kv cache
+#[derive(Debug, Eq, PartialEq, thiserror::Error)]
+pub enum KvCacheConversionError {
+    /// Sequence id conversion to i32 failed
+    #[error("Provided sequence id is too large for a i32")]
+    SeqIdTooLarge(#[source] TryFromIntError),
+    /// Position 0 conversion to i32 failed
+    #[error("Provided start position is too large for a i32")]
+    P0TooLarge(#[source] TryFromIntError),
+    /// Position 1 conversion to i32 failed
+    #[error("Provided end position is too large for a i32")]
+    P1TooLarge(#[source] TryFromIntError),
+}
 
 impl LlamaContext<'_> {
     /// Copy the cache from one sequence to another.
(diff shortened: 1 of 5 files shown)


llama-cpp-2/src/context/kv_cache.rs
```diff
@@ -121,6 +121,19 @@ impl LlamaContext<'_> {
...
+    /// Remove all tokens within the specified range `[p0, p1)` from the KV cache
+    /// Returns `false` only when partial sequence removals fail. Full sequence removals always succeed.
+    ///
+    /// # Parameters
+    ///
+    /// * `seq_id` - The sequence id to remove the tokens from. If < 0, matches all sequences
+    /// * `p0` - The start position of the cache to remove. If < 0, the entire cache is removed up to `p1`
+    /// * `p1` - The end position of the cache to remove. If < 0, the entire cache is removed starting from `p0`
+    #[must_use]
+    pub fn kv_cache_seq_rm(&mut self, seq_id: i32, p0: i32, p1: i32) -> bool {
```
MarcusDunn, line 133: I would prefer for these (unsure how consistently this is done elsewhere) that we don't have special behaviour for negative ints (however well documented it is). An Option<NonZeroU16> is more clear.

MarcusDunn (commented):
Looks good. I'd like a signature change to reflect how negative numbers work. This is a great addition.

llama-cpp-2/src/context/kv_cache.rs
brittlewis12: that's a great call!

would you like me to apply this change to the `seq_id` as well? it shares the same "-1 means all" semantics, tho I see in other kv cache-related methods so far we've left that an `i32`

---

by the way, i don't have a use case for it today, but there are already models with context lengths greater than 65k, so `u16` for `llama_pos` can't fully express the range of valid values for such large models.. but of course jumping to a `u32` introduces the overflow problem on the downcast to `i32` for values between 2B & 4B, so it's probably a fine tradeoff to opt for simplicity & leave as `u16`s for now.

alternatively, we could accept `u32` for `llama_pos` & return a result if `i32::try_from` fails, rather than a bool, which might be a bit more rust-y anyhow?

brittlewis12: for example:
```rust
    /// Remove all tokens within the specified range `[p0, p1)` from the KV cache
    /// Returns `false` only when partial sequence removals fail. Full sequence removals always succeed.
    ///
    /// # Returns
    /// A `Result` indicating whether the operation was successful. If the sequence id or position
    /// exceed the maximum i32 value, no remonval is attempted and an `Err` is returned.
    ///
    /// # Parameters
    ///
    /// * `seq_id` - The sequence id to remove the tokens from. If `None`, matches all sequences
    /// * `p0` - The start position of the cache to remove. If `None`, the entire cache is removed up to `p1`
    /// * `p1` - The end position of the cache to remove. If `None`, the entire cache is removed starting from `p0`
    #[must_use]
    pub fn kv_cache_seq_rm(
        &mut self,
        seq_id: Option<u32>,
        p0: Option<u32>,
        p1: Option<u32>,
    ) -> Result<bool, TryFromIntError> {
        let seq_id = seq_id.map_or(Ok(-1), i32::try_from)?;
        let p0 = p0.map_or(Ok(-1), i32::try_from)?;
        let p1 = p1.map_or(Ok(-1), i32::try_from)?;

        Ok(
            unsafe {
                llama_cpp_sys_2::llama_kv_cache_seq_rm(self.context.as_ptr(), seq_id, p0, p1)
            },
        )
    }
```

MarcusDunn: that looks great.

MarcusDunn: and feel free to make similar changes elsewhere.

brittlewis12: great! do you mind bringing `std::num::TryFromIntError` into the crate's public API, or would you prefer something using `this_error`, like maybe:
```rust
/// Errors that can occur when attempting to manipulate the kv cache
#[derive(Debug, Eq, PartialEq, thiserror::Error)]
pub enum KvCacheConversionError {
    /// Sequence id conversion to i32 failed
    #[error("Provided sequence id is too large for a i32")]
    SeqIdTooLarge(#[source] TryFromIntError),
    /// Position 0 conversion to i32 failed
    #[error("Provided start position is too large for a i32")]
    P0TooLarge(#[source] TryFromIntError),
    /// Position 1 conversion to i32 failed
    #[error("Provided end position is too large for a i32")]
    P1TooLarge(#[source] TryFromIntError),
}
```

brittlewis12: ok, [ran with something like this](https://github.com/utilityai/llama-cpp-rs/pull/510/commits/7d1b2d53e2b677420539a339d4e9b522a206427b), but let me know if you'd prefer something else!

MarcusDunn (approved):

MarcusDunn:
Looks good. If Linux tests and Mac build pass I'll merge.

PR #517, opened 2024-10-07
ci: fix version bumping GitHub Actions job

babichjacob:
examples were placed in the `examples` directory in https://github.com/utilityai/llama-cpp-rs/pull/482 but this change was missed, so there's no longer automatic version bumping that might be contributing to the crates.io releases being behind

P.S. I've been enjoying how [Release-plz](https://release-plz.ieni.dev/) works if you'd find it beneficial to replace the `publish-upon-release` and `update-toml-version` workflows with it. (I prefer [the GitHub app approach](https://release-plz.ieni.dev/docs/github/token#use-a-github-app))


babichjacob:
requires #519 actually, oops

MarcusDunn (approved):

PR #521, opened 2024-10-08
fix: make `llama-cpp-sys-2` publishable again

### LINKED ISSUE #518: crates.io publish jobs are failing even though it builds successfully (locally / for me)
https://github.com/utilityai/llama-cpp-rs/actions/runs/10710926865/job/29698650066 failed

It claimed
>   CMake Error: The source directory "/home/runner/work/llama-cpp-rs/llama-cpp-rs/target/package/llama-cpp-sys-2-0.1.69/target/debug/build/llama-cpp-sys-2-d4132d04ad575bbe/out/llama.cpp" does not appear to contain CMakeLists.txt.

even though locally it is there. It's been a few weeks and there have been some changes. so maybe this was a fluke and it might just need to be attempted again? Though, the `version` specified in `llama-cpp-2` and `llama-cpp-sys-2` needs to be manually bumped because the job dedicated to it hasn't been doing it (#517) and 0.1.69 is already published. Possibly just fast forward to `0.1.80` as there are already GitHub releases up to `0.1.79`?

babichjacob:
~~more inclusions are likely needed for CUDA and Vulkan,~~ and the others

both Vulkan and CUDA (I don't have the CUDA Toolkit installed so I don't even know how this is possible) compile when [downloading it back from crates.io](https://crates.io/crates/babichjacob-llama-cpp-sys-2) on my end, but I'm not sure I did that right... 

closes #518

### DIFF
--- a/llama-cpp-sys-2/Cargo.toml
+++ b/llama-cpp-sys-2/Cargo.toml
@@ -11,46 +11,46 @@ include = [
     "wrapper.h",
     "build.rs",
     "/src",
-    "/llama.cpp/ggml/src/ggml.c",
-    "/llama.cpp/ggml/include/ggml.h",
-    "/llama.cpp/ggml/src/ggml-alloc.c",
-    "/llama.cpp/ggml/include/ggml-alloc.h",
-    "/llama.cpp/ggml/src/ggml-backend.c",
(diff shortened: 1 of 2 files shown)


MarcusDunn:
thanks for the PR. 0.1.81 is up!

PR #561, opened 2024-10-28
build.rs: also search for library files in lib64/

AsbjornOlling:
This fixes compilation on my machine and in Nix.

It seems like `libllama.a` ends up in `${out_dir}/lib64/`, and not in `${out_dir}/lib/`, like the build script assumes.

Why? I'm not quite sure. But it hapens on the current nixos-unstable, using cargo, cmake and clang from nixpkgs.

It also affects building in the nix build sandbox (using e.g. `nixpkgs.rustPlatofrm.buildRustPackage`), so after this patch it is again possible to build rust applications that depend on this library using the nix build system.

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -65,7 +65,7 @@ fn extract_lib_names(out_dir: &Path, build_shared_libs: bool) -> Vec<String> {
             "*.a"
         }
     };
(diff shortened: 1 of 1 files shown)


MarcusDunn:
Bonus points to add an error message to the assert_eq. Otherwise looks great. Thanks for the PR. I will merge once tests look mostly green.

PR #579, opened 2024-11-21
wrap llama_batch_get_one

tinglou:
wrap llama_cpp_sys_2::llama_batch_get_one and safer drop

### DIFF
--- a/llama-cpp-2/src/llama_batch.rs
+++ b/llama-cpp-2/src/llama_batch.rs
@@ -1,6 +1,6 @@
 //! Safe wrapper around `llama_batch`.
 
-use crate::token::{self, LlamaToken};
+use crate::token::LlamaToken;
 use llama_cpp_sys_2::{llama_batch, llama_batch_free, llama_batch_init, llama_pos, llama_seq_id};
 
 /// A safe wrapper around `llama_batch`.
--- a/llama-cpp-2/src/llama_batch.rs
+++ b/llama-cpp-2/src/llama_batch.rs
@@ -149,6 +149,25 @@ impl LlamaBatch {
         }
     }
 
+    /// llama_batch_get_one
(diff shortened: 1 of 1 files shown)


llama-cpp-2/src/llama_batch.rs
```diff
@@ -149,6 +149,25 @@ impl LlamaBatch {
...
     }
 
+    /// llama_batch_get_one
+    /// Return batch for single sequence of tokens starting at pos_0
+    ///
+    /// NOTE: this is a helper function to facilitate transition to the new batch API
+    ///
+    pub fn get_one(tokens: &[LlamaToken], pos_0: llama_pos, seq_id: llama_seq_id) -> Self {
+        unsafe {
+            let ptr = tokens.as_ptr() as *mut i32;
```
MarcusDunn, line 159: Why does this need to be mut? If it doesn't, can you make a PR to llama-cpp instead of casting?

Also what are the implications of an empty slice here?

tinglou:
```
#[repr(C)]
#[derive(Debug, Copy, Clone, PartialEq)]
pub struct llama_batch {
    pub n_tokens: i32,
    pub token: *mut llama_token,
    pub embd: *mut f32,
    pub pos: *mut llama_pos,
    pub n_seq_id: *mut i32,
    pub seq_id: *mut *mut llama_seq_id,
    pub logits: *mut i8,
    pub all_pos_0: llama_pos,
    pub all_pos_1: llama_pos,
    pub all_seq_id: llama_seq_id,
}
```
The orignal field `token` is mut.

Yes, the slice should not be empty.

MarcusDunn (approved):

MarcusDunn:
included in latest, thanks for the PR

Issue #548, opened 2024-10-21
Adopting upstream sampling API refactors


brittlewis12:
Llama.cpp implemented breaking [sampling API refactors](https://github.com/ggerganov/llama.cpp/pull/9294/files#top) which changed from exposing stateless sampling functions for use in the generation loop to a stateful sampler chain which is setup up front and can be manipulated from there.

I [took a stab](https://github.com/utilityai/llama-cpp-rs/compare/main...brittlewis12:adopt-updated-sampling-api?expand=1) at spiking out how integrating the new API into llama-cpp-rs could look, but it's far from releasable. I stopped once I got `simple` running successfully.

Would love your perspective on approaches to unifying the existing sampling interfaces exposed by the crate, what to do about llama token data array sampling functions, handling grammar sampling, as well as feedback on the rust seams over llama.cpp's latest sampling APIs.

Once we're on the same page, I can flesh out last missing pieces like sampler timings & any higher level abstractions over sampling parameters — it seems the library doesn't expose any sampling parameter structs directly any more :/

I have these changes up as a branch on my fork, but if you'd prefer I'm more than happy to collaborate on a proper draft pull request.

MarcusDunn:
on our end, we've stopped using the `llama-cpp-rs` sampling API, I kept it around because removing it was more work.

The closer we can be to *safe direct bindings* to upstream the happier I am, ergonomics is secondary. I like the direction the PR is going (wrapping the `llama.cpp` structs and their calls).

If you want to get it into a minimal state you would be happy to use I can do a more detailed review (it looks pretty close). No need to cover 100% of the API - just whatever you find useful.

eugenehp:
Here's an attempt on adding the sampler. 
Still more wip on common sampler and grammar, but basic pieces are there.
https://github.com/eugenehp/llama-cpp-rs/tree/sampler

MarcusDunn:
fixed in #580

PR #580, opened 2024-11-26
Implement new sampler API and bump llama.cpp

volesen:
`llama.cpp` introduces a new sampling API in b3680. This PR aims to replace the previous custom sampling API, with a wrapper around the new `llama.cpp` sampling abstraction.

We remove the custom sampling functions from `LlamaContext`, *so this is a breaking change*.

Furthermore, we bump `llama.cpp` to b3750, which finalizes the new `llama_perf` features, so we can more easily preserve the timings API.

Credits to @brittlewis12, whose draft we copied in large parts. (#548)

It's probably worth it to squash commits before merging.

The new sampler is showcased in the `simple` example and  can be used like this:

```rs
let sampler_params = LlamaSamplerChainParams::default();
let mut sampler = LlamaSampler::new(sampler_params)?.add_dist(seed.unwrap_or(1234));
let new_token = sampler.sample(&ctx, batch.n_tokens() - 1);
sampler.accept(token);
```


MarcusDunn:
awesome work! I'll get this in + release soon as tests look good.

PR #584, opened 2024-12-02
Performacne: Cast Vec<LlamaToken> from/to Vec<i32> without memory copy

tinglou:
Add helper methods for casting Vec<LlamaToken> from/to Vec<i32> without memory copy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,9 +146,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.22"
+version = "4.5.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69371e34337c4c984bbe322360c2547210bf632eb2814bbe78a6e87a2935bd2b"
+checksum = "3135e7ec2ef7b10c6ed8950f0f792ed96ee093fa088608f1c76e569722700c84"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -156,9 +156,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.22"
+version = "4.5.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e24c1b4099818523236a8ca881d2b45db98dadfb4625cf6608c12069fcbbde1"
+checksum = "30582fc632330df2bd26877bde0c1f4470d57c582bbc070376afcd04d8cb4838"
 dependencies = [
  "anstream",
  "anstyle",
@@ -180,9 +180,9 @@ dependencies = [
 
 [[package]]
 name = "clap_lex"
-version = "0.7.1"
+version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4b82cf0babdbd58558212896d1a4272303a57bdb245c2bf1147185fb45640e70"
+checksum = "f46ad14479a25103f283c0f10005961cf086d8dc42205bb44c46ac563475dca6"
 
 [[package]]
 name = "cmake"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ pprof = "0.13.0"
 bindgen = "0.69.5"
 cc = "1.2.3"
 anyhow = "1.0.94"
-clap = "4.5.22"
+clap = "4.5.23"
 encoding_rs = "0.8.35"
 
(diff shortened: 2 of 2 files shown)


MarcusDunn:
I'd have to see profiling showing an improvement before accepting this. I would be somewhat surprised if rustc is unable to optimize this out without unsafe.

MarcusDunn:
https://godbolt.org/z/9Gs8z3bxE

Not representative of a whole program, but in the simple case, rustc figures it out.

tinglou:
> I'd like to see profiling showing an improvement before accepting this. I would be somewhat surprised if rustc is unable to optimize this out.

For example, `collect` in the following  will call `Vec::from_iter`
```
        Ok(buffer.into_iter().map(LlamaToken).collect())
```

And check the implemenation is std:
```
impl<T> SpecFromIter<T, IntoIter<T>> for Vec<T> {
    fn from_iter(iterator: IntoIter<T>) -> Self {
        // A common case is passing a vector into a function which immediately
        // re-collects into a vector. We can short circuit this if the IntoIter
        // has not been advanced at all.
        // When it has been advanced We can also reuse the memory and move the data to the front.
        // But we only do so when the resulting Vec wouldn't have more unused capacity
        // than creating it through the generic FromIterator implementation would. That limitation
        // is not strictly necessary as Vec's allocation behavior is intentionally unspecified.
        // But it is a conservative choice.
        let has_advanced = iterator.buf != iterator.ptr;
        if !has_advanced || iterator.len() >= iterator.cap / 2 {
            unsafe {
                let it = ManuallyDrop::new(iterator);
                if has_advanced {
                    ptr::copy(it.ptr.as_ptr(), it.buf.as_ptr(), it.len());
                }
                return Vec::from_nonnull(it.buf, it.len(), it.cap);
            }
        }

        let mut vec = Vec::new();
        // must delegate to spec_extend() since extend() itself delegates
        // to spec_from for empty Vecs
        vec.spec_extend(iterator);
        vec
    }
}
```
The function create a new `Vec` and allocate memory for it.

MarcusDunn:
Reasoning about compiler optimizations by looking at source code ends up being a guessing game from my experience. Because this code is unsafe, the bar for accepting it is going to be higher than a non-unsafe optimization.

I'm going to either need
- profiling
- godbolt examples (with optimizations on)

That show an improvement.

nkoppel:
Yeah, it seems like rustc is able to optimize that away. See that [this godbolt example](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:12,endLineNumber:4,positionColumn:12,positionLineNumber:4,selectionStartColumn:12,selectionStartLineNumber:4,startColumn:12,startLineNumber:4),source:'%23%5Brepr(transparent)%5D%0Apub+struct+Token(i32)%3B%0A%0A%23%5Bno_mangle%5D%0Afn+convert1(tokens:+Vec%3Ci32%3E)+-%3E+Vec%3CToken%3E+%7B%0A++++tokens.into_iter().map(Token).collect()%0A%7D'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1830,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-O',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+rustc+1.83.0+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'rustc+1.83.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.83.0+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) does not produce any loops or allocater calls.

PR #594, opened 2024-12-04
Add sampling API back to LlamaTokenDataArray; Add DRY and XTC Samplers

nkoppel:
The changes in 0.1.84 removed the ability to use samplers with `LlamaTokenDataArray`s. These changes add the sampling methods of version 0.1.83 back to `LlamaTokenDataArray`, adds support for the XTC and DRY samplers, and allows users to apply a `LlamaSampler` to a `LlamaTokenDataArray`.

These changes enable users to mix custom samplers with llama.cpp samplers, and makes the `LlamaTokenDataArray` struct useful again.

Let me know if you'd like to see any changes!


llama-cpp-2/src/token/data_array.rs
```diff
@@ -1,18 +1,24 @@
...
+use super::LlamaToken;
 
 /// a safe wrapper around `llama_token_data_array`.
 #[derive(Debug, Clone, PartialEq)]
 #[allow(clippy::module_name_repetitions)]
 pub struct LlamaTokenDataArray {
     /// the underlying data
     pub data: Vec<LlamaTokenData>,
+    /// the selected token
+    pub selected: i64,
```
MarcusDunn, line 15: I think I would prefer a Option<u32> or something similar. I would prefer not to use -1 as a sentinel in a pub struct like this. (are other negative values valid?)

llama-cpp-2/src/token/data_array.rs
```diff
@@ -47,4 +57,320 @@ impl LlamaTokenDataArray {
...
+    /// SAFETY:
+    /// [modify] cannot change the data pointer.
+    /// if the data is not sorted, sorted must be false.
+    /// the size of the data can only decrease (i.e you cannot add new elements).
+    pub(crate) unsafe fn modify_as_c_llama_token_data_array<T>(
+        &mut self,
+        modify: impl FnOnce(&mut llama_cpp_sys_2::llama_token_data_array) -> T,
+    ) -> T {
+        let size = self.data.len();
+        let data = self.data.as_mut_ptr().cast();
```
MarcusDunn, line 85: unless there are platform considerations, I think explicitly mentioning the type to cast to would be best here.

llama-cpp-2/src/token/data_array.rs
```diff
@@ -47,4 +57,320 @@ impl LlamaTokenDataArray {
...
+    }
+
+    pub(crate) unsafe fn apply_and_free_sampler(
+        &mut self,
+        sampler_fn: impl FnOnce() -> *mut llama_cpp_sys_2::llama_sampler,
+    ) {
+        let sampler = sampler_fn();
+        self.apply_sampler(sampler);
+        llama_cpp_sys_2::llama_sampler_free(sampler);
+    }
```
MarcusDunn, line 117: I know this is pub(crate) but I think a repr(transparent) struct with `Drop` would make maintaining this abstraction easier.

MarcusDunn (commented):
This looks good. I've got a couple of comments that don't _require_ changes, but I would like your thoughts on before I merge.

llama-cpp-2/src/token/data_array.rs
nkoppel: It is now an `Option<usize>`, which should be much clearer as to its purpose and easier to use.

llama-cpp-2/src/token/data_array.rs
nkoppel: Done!

nkoppel:
I've gone ahead and overhauled the sampling API in a manner which resolves your third review comment, but which is very different from what came before. In my latest version, `LlamaSampler`s can be defined as either a chain or a single sampler as determined by `LlamaSamplerParams`, which is an enum struct that contains all parameters needed to construct a `LlamaSampler`. I used this to factor out all of the `LlamaSampler::add_` and `LlamaTokenDataArray::sample_` methods in favor of specifying which sampler to use using the enum. This deduplicates a lot of code and does not make the interface much more clunky than it was.

If you think that this API is too far from the raw API or complicates things too much, let me know.

llama-cpp-2/src/sampling.rs
```diff
@@ -21,236 +21,155 @@ impl Debug for LlamaSampler {
...
-
-    /// Minimum P sampling as described in <https://github.com/ggerganov/llama.cpp/pull/3841>
-    #[must_use]
-    #[allow(unused_mut)]
-    pub fn add_min_p(mut self, p: f32, min_keep: usize) -> Self {
-        unsafe {
-            let min_p_sampler = llama_cpp_sys_2::llama_sampler_init_min_p(p, min_keep);
-            llama_cpp_sys_2::llama_sampler_chain_add(self.sampler.as_ptr(), min_p_sampler);
-        }
+unsafe fn new_inner(params: LlamaSamplerParams) -> *mut llama_cpp_sys_2::llama_sampler {
```
MarcusDunn, line 24: I don't think this has to be marked as unsafe. Are there invariants callers need to uphold? If so, document them.

llama-cpp-2/src/sampling.rs
```diff
@@ -3,16 +3,16 @@ pub mod params;
...
 use crate::context::LlamaContext;
-use crate::model::LlamaModel;
+use crate::token::data_array::LlamaTokenDataArray;
 use crate::token::LlamaToken;
-use crate::LlamaSamplerError;
+
+use params::LlamaSamplerParams;
 
 /// A safe wrapper around `llama_sampler`.
 pub struct LlamaSampler {
```
MarcusDunn, line 14: Does this require a lifetime given there's one in LlamaSamplerParams?

If we drop a model after creating one of these I think it could lead to a use after free from CPP?

MarcusDunn (changes requested):
I think there's a bit of thinking to do around lifetimes before this gets merged. 

If you want to remove the samplers that require lifetime logic, it would be nice to get a first iteration of the trivial samplers in, then we can add the more complex ones in a separate PR. There's a lot of great work here I'd hate to see not in main because we're battling encoding model lifetimes guarantees into the sampler.

MarcusDunn:
I don't mind the new API. I weakly preferred the "closer to direct calls" one.

MarcusDunn:
I think a LlamaSampler struct with a collection of factory functions would be preferable to the params + enum approach.

llama-cpp-2/src/sampling.rs
nkoppel: No, and the lifetime situation is no different than it was before. For example, constructing a grammar sampler uses the model's tokenizer and the grammar strings to initialize itself, and then owns all of its data, so we do not need to keep the data and model alive to use the sampler. The lifetime on `LlamaSamplerParams` is a little ugly, but is necessary to include model references and makes syntax simpler in the other places it is used.

nkoppel:
Okay, I've rewritten the API again, and I think that this is about as simple as it will get while still allowing LlamaSampler to be either a chain or a single sampler. It is also about as close to the raw api as it was before this pr. I will write documentation for the new methods on LlamaSampler later tody.

llama-cpp-2/src/sampling.rs
MarcusDunn: This makes total sense. Thank you.

llama-cpp-2/src/sampling.rs
```diff
@@ -22,187 +20,157 @@ impl Debug for LlamaSampler {
...
-            let dist_sampler = llama_cpp_sys_2::llama_sampler_init_dist(seed);
-            llama_cpp_sys_2::llama_sampler_chain_add(self.sampler.as_ptr(), dist_sampler);
+    /// Accepts several tokens from the sampler or context, possibly updating the internal state of
+    /// certain samplers (e.g. grammar, repetition, etc.)
+    pub fn accept_many(&mut self, tokens: &[LlamaToken]) {
+        for token in tokens {
+            unsafe { llama_cpp_sys_2::llama_sampler_accept(self.sampler, token.0) }
         }
-
-        self
```
MarcusDunn, line 60: `impl IntoIter` > slice.

https://rust-lang.github.io/api-guidelines/flexibility.html#functions-minimize-assumptions-about-parameters-by-using-generics-c-generic

llama-cpp-2/src/sampling.rs
```diff
@@ -22,187 +20,157 @@ impl Debug for LlamaSampler {
...
+    /// certain samplers (e.g. grammar, repetition, etc.)
     #[must_use]
-    #[allow(unused_mut)]
-    pub fn add_top_k(mut self, k: i32) -> Self {
-        unsafe {
-            let top_k_sampler = llama_cpp_sys_2::llama_sampler_init_top_k(k);
-            llama_cpp_sys_2::llama_sampler_chain_add(self.sampler.as_ptr(), top_k_sampler);
-        }
-
+    pub fn with_tokens(mut self, tokens: &[LlamaToken]) -> Self {
```
MarcusDunn, line 55: same here about `impl IntoIter`

llama-cpp-2/src/sampling.rs
```diff
@@ -213,8 +181,8 @@ impl LlamaSampler {
         penalize_nl: bool,
         ignore_eos: bool,
```
MarcusDunn, line 182: nit: these may be worth turning into enums. Call sites like 

```rust
Self::penalties(
            model.n_vocab(),
            model.token_eos().0,
            model.token_nl().0,
            penalty_last_n,
            penalty_repeat,
            penalty_freq,
            penalty_present,
            false,
            true,
        )
```

are hard to figure out what true and false mean

llama-cpp-2/src/token/data_array.rs
```diff
@@ -47,4 +60,89 @@ impl LlamaTokenDataArray {
...
+            );
+        }
+        self.data.set_len(c_llama_token_data_array.size);
+
+        self.sorted = c_llama_token_data_array.sorted;
+        self.selected = c_llama_token_data_array
+            .selected
+            .try_into()
+            .ok()
+            .filter(|&s| s < self.data.len());
```
MarcusDunn, line 121: when does this filter apply?

MarcusDunn (approved):
minor nitpicks. This looks good - address / dismiss as you sit fit and once you're happy I will merge long as tests pass.

I think this is also one of the parts of the library that can be reasonably tested (stuff that does inference or loads a model is too hard), a couple tests (preferably in docs) would be lovely.

Thanks for the great PR and patience with the back and forth.

llama-cpp-2/src/token/data_array.rs
nkoppel: This is just to ensure that we don't store a value in `selected` that is out of bounds for the array. This will probably only occur if the user applies a selection operation (greedy, dist, mirostat) before applying a filter, which seems unlikely. In the meantime, it make the library more robust in theorey, so I think it is worth keeping.

llama-cpp-2/src/sampling.rs
nkoppel: I expect this method to be used very rarely compared to `penalties_simple`, so I'm not sure that creating two public enums is worth it for this.

PR #598, opened 2024-12-08
Added Reranker example

srv1n:
The parent llama.cpp repo recently added support for reranking. 

- https://github.com/ggerganov/llama.cpp/pull/9510

This PR 
- Adds a rerank example
- adds a new pooling type called rank( matches with llama.cpp implementation) in params
- initialized_logits made public in llamabatch 

Validated against all examples in the llama.cpp and BGE-reranker-v2-m3 and all rerank scores match. 

Please review, make edits and feel free to commit. Thank you for the awesome repo!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -935,6 +935,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7a66a03ae7c801facd77a29370b4faec201768915ac14a721ba36f20bc9c209b"
 
 [[package]]
+name = "reranker"
+version = "0.1.86"
+dependencies = [
+ "anyhow",
+ "clap",
+ "encoding_rs",
+ "hf-hub",
+ "llama-cpp-2",
+]
+
+[[package]]
 name = "ring"
 version = "0.17.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


MarcusDunn (commented):
This looks great! Thanks for the PR.

my only question is, why did you make initialized_logits public?

MarcusDunn:
also, is the submodule up to date enough to support this?

srv1n:
I was trying to match it as closely with the original llama.cpp repo. Specifically the batch decode treats [pooling type none and the other pooling types differently. ](https://github.com/ggerganov/llama.cpp/blob/ce8784bdb153ff7794dde5a50b0ebfa51baa6171/examples/embedding/embedding.cpp#L55C1-L79C2). 

I initially started off trying to modify the embedding example to match the original repo ( since reranking in llama.cpp is built into the examples/embedding.cpp ). 

Its actually not being used in the reranking (since pooling is set to rank), but does it make sense to leave it public so we could use it with pooling type none if required?

MarcusDunn:
sorry for the late reply!

I would prefer we keep it private unless it is required to have a feature work.

srv1n:
My bad finally got around to it.  Updated now. initialized_logits  is not public anymore.

PR #605, opened 2024-12-19
Build for aarch64-linux-android

AsbjornOlling:
...does what it says on the tin.

We're building a Godot plugin, and people are requesting android support.

It seems to work for us:
https://github.com/nobodywho-ooo/nobodywho/pull/67

There are two changes:
  1. I add the cmake flags suggested in the llama.cpp docs [here](https://github.com/ggerganov/llama.cpp/blob/master/docs/android.md), if the target string contains "android" and "aarch64"
  2. The DRY sampler contains an `i8` type that is an `u8` in the android NDK. I just set this type to be a `u8` whe compiling for android.

I'm not really an android dev, so not an expert on any of this. But it seems to work.

### DIFF
--- a/llama-cpp-2/src/sampling.rs
+++ b/llama-cpp-2/src/sampling.rs
@@ -20,6 +20,15 @@ impl Debug for LlamaSampler {
     }
 }
 
+// this is needed for the dry sampler to typecheck on android
+// ...because what is normally an i8, is an u8 
+#[cfg(target_os = "android")]
+type CChar = u8;
+
+#[cfg(not(target_os = "android"))]
+type CChar = i8;
+
+
 impl LlamaSampler {
     /// Sample and accept a token from the idx-th output of the last evaluation
     #[must_use]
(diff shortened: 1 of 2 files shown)


MarcusDunn (approved):
thanks for the PR. LGTM.

llama-cpp-2/src/sampling.rs
```diff
@@ -254,7 +263,7 @@ impl LlamaSampler {
             .into_iter()
             .map(|s| CString::new(s.as_ref()).unwrap())
             .collect();
-        let mut seq_breaker_pointers: Vec<*const i8> =
+        let mut seq_breaker_pointers: Vec<*const CChar> =
```
MarcusDunn, line 266:
```suggestion
        let mut seq_breaker_pointers =
```

llama-cpp-2/src/sampling.rs
```diff
@@ -20,6 +20,15 @@ impl Debug for LlamaSampler {
...
 
+// this is needed for the dry sampler to typecheck on android
+// ...because what is normally an i8, is an u8 
+#[cfg(target_os = "android")]
+type CChar = u8;
+
+#[cfg(not(target_os = "android"))]
+type CChar = i8;
+
+
```
MarcusDunn, line 31:
```suggestion
// this is needed for the dry sampler to typecheck on android
// ...because what is normally an i8, is an u8 
#[cfg(target_os = "android")]
type CChar = u8;

#[cfg(not(target_os = "android"))]
type CChar = i8;


```

MarcusDunn:
If the compiler can infer the type (and we avoid the platform-specific-types) that would be best.

bindgen often also generates types that work accross platforms, unsure what it should be in this case.

PR #622, opened 2025-01-21
Update llama.cpp dependency to the latest version; Fix all clippy lints.

nkoppel:
I updated the llama.cpp dependency to latest as of today in order to run the new [Deepseek R1 models](https://github.com/deepseek-ai/DeepSeek-R1) with this library. I also fixed all clippy lints while I was at it, because it annoyed me that they exist.

The only break to the user-facing API is that `LlamaSampler::penalties_simple` has been removed, and `LlamaSampler::penalties` has had a number of parameters removed, because these parameters were removed upstream in llama.cpp. Let me know if you want to see any changes!


MarcusDunn:
This is great! Thanks for the PR!

MarcusDunn (approved):

PR #629, opened 2025-01-29
Fix Android build

vlovich:
The standard library predefines c_char exactly for this purpose to get rid of manual CChar type aliases. While the alias in theory looks fine, it's actually not if building for x64 Android. Similar problems exist in the CMake setup around assuming Android only targets aarch64. There's also a problem where openmp is enabled by default but for me on Android this just results in a bunch of undefined OpenMP symbols. Finally, I had issues linking against the C++ stdlib statically so made a default feature that links to the dynamic C++ library. I have confirmed that a complex Android project builds and links under cargo-ndk for all the triples listed by the cargo-ndk project.

### DIFF
--- a/llama-cpp-2/Cargo.toml
+++ b/llama-cpp-2/Cargo.toml
@@ -18,7 +18,7 @@ tracing = { workspace = true }
 encoding_rs = { workspace = true }
 
 [features]
-default = ["openmp"]
+default = ["openmp", "android-shared-stdcxx"]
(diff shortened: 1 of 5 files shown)


MarcusDunn (approved):

vlovich:
I don't believe the build failures have anything to do with my change. Looks like some platforms have bitrotted?

MarcusDunn:
Yeah they have, they've been failing for a while now. Thanks for the PR!

Issue #631, opened 2025-01-31
Windows build broken? error: could not find native static library `ggml-base`, perhaps an -L flag is missing?


vlovich:
Not sure what's going on but for some reason getting this error on both windows native & linux->windows cross-compile builds. Is this something you're seeing too? Can't think what changed.

babichjacob:
I'm on Windows x86_64 (AMD Ryzen 5 5600) and not getting any errors compiling this repository after a fresh `git clone https://github.com/utilityai/llama-cpp-rs --recursive` nor getting any at runtime, nor when adding `llama-cpp-2`'s most recent published version (`0.1.91`) as a dependency to another crate and building there.

Is there something I missed trying to reproduce?

vlovich:
Coworker reported the same issue on Windows natively but I can't speak to their setup as I haven't had this issue. For me it happened trying to cross-compile to Windows from Linux:

```
rustup target add x86_64-pc-windows-gnu
RUSTC_WRAPPER= cargo build --target "x86_64-pc-windows-gnu" --verbose
```

vlovich:
I have a hunch that this has something to do with mingw builds. I believe my coworker was getting the mingw compiler as their C++ compiler for whatever reason.

PR #632, opened 2025-01-31
Fix builds for cross-compilation with mingw

vlovich:
I've confirmed this fixes the issue cross-compiling to Windows from Linux. I have a coworker who says their Windows native build is failing although in theory that shouldn't be this patch. Will investigate further next week but figured I'd put up a PR for discussion.

#631


babichjacob:
Per https://github.com/utilityai/llama-cpp-rs/issues/631#issuecomment-2628082433 my build wasn't broken to begin with, but you're right, my native Windows (AMD64) build doesn't start failing with this PR either.

MarcusDunn (approved):

MarcusDunn:
Thanks for the PR. I do not use windows, but if you want to setup a CI pipeline that will be the best way to make sure this stays fixed. (this is far from the first time windows support has been broken).

PR #639, opened 2025-02-10
Fix packaging

vlovich:
Comment describes what broke & what cleanup needs to happen upstream to remove the need for this change. Another alternative, if we do want to include the generated build-info.cpp after all is to pass `--allow-untracked` to the publish command but that feels weird because where build-info.cpp is getting generated is an "unstable" decision internal to the llama.cpp project.

### DIFF
--- a/llama-cpp-sys-2/Cargo.toml
+++ b/llama-cpp-sys-2/Cargo.toml
@@ -24,6 +24,17 @@ include = [
 
     "/llama.cpp/convert_hf_to_gguf.py", # Yes, it's required
 
+    # Erroneously the llama.cpp code currently generates the build-info.cpp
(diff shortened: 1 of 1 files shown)


MarcusDunn (approved):

MarcusDunn:
seems this does not quite get us there.

https://github.com/utilityai/llama-cpp-rs/actions/runs/13245909131/job/36972240379

MarcusDunn:
I'm able to reproduce locally with `cargo publish --dry-run -p llama-cpp-sys-2`

vlovich:
Yup I put up a PR fixing the issue. Not sure how I missed it because in
principle I had been verifying the command. Verified the command works on
my latest PR

PR #647, opened 2025-02-10
Fix llama.cpp build dirtying src and breaking packaging

vlovich:
Move the build-info.cpp to the output target after building to avoid polluting during packaging and causing verification to fail.

Turns out I didn't actually test dry-run the final version of the PR I pushed? Not sure what happened. This passes.


MarcusDunn:
works! 

https://crates.io/crates/llama-cpp-2/0.1.93

PR #671, opened 2025-02-22
Don't imply dynamic llama.cpp just because CUDA is on

vlovich:
Link against CUDA statically as well to maintain consistency with GGML_STATIC although technically that's our discretion. Verified that now just enabling CUDA still links llama.cpp statically (as well as all the supporting CUDA libraries which simplifies distribution).

Verified this works on Linux. Have not verified on Windows yet.


MarcusDunn (approved):

MarcusDunn:
did you intent to also modify the submodule?

Looks good otherwise.

PR #677, opened 2025-02-24
Make LogOptions public

kusaanko:
I can't construct LogOptions because its members are private.
In order to use `send_logs_to_tracing`, those are needed to public.
This pull request fixes it.


MarcusDunn:
You can create log options using `LogOptions::default()` I would prefer not to have the struct be all public so we can add fields in a backwards comparable way. Documentation on the struct with an example on how to create it to prevent future confusion would be appreciated.

kusaanko:
Oh I forgot ::default func. I have thought that Default can be only used in

```rust
LogOptions {
    ..Default::default()
}
```

Thank you!

PR #699, opened 2025-03-25
Reimplement get_chat_template to use `llama_model_chat_template`

AsbjornOlling:
Adds support for the new `llama_model_chat_template` name parameter, to allow getting a chat template by name. This was added to llama.cpp in b4524.

`get_chat_template` now takes an `Option<&str>`. If None, it returns the default template, if Some, it returns the template by that name.

This is a *breaking change* since it changes the type signature of the `get_chat_template` function.

The simple example runs and works, as do the unit tests- but none of these cover the functionality of `get_chat_template`. To ensure that this code works, I have tested it with the unit testing suite [on this project](https://github.com/nobodywho-ooo/nobodywho/).

Some alternatives to consider:
- we could maintain backwards compatibility by keeping the type signature of `get_chat_template`, and introduce a new method which takes the name parameter. maybe `get_chat_template_by_name`? It would still technically be a breaking change because of the new variants in the error type, but a smaller one.
- we could use a custom enum instead of `Option<&str>`, with more descriptive names. It could have variants like `ChatTemplateName::Default`, `ChatTemplateName::ToolUse`, and `ChatTempalteName::Custom(&str)`
- the old solution used `meta_val_str` to get the template, with a bunch of ceremony around it. If there is a reason to do it in that more complicated way, I didn't understand why. If anyone knows a benefit, we could consider keeping the old implementation.
- ~~The method could be called `chat_template` for consistency, instead of `get_chat_template`, since it wraps a function named `llama_model_chat_template`. It seems all other methods that begin with `llama_model` just get that prefix stripped for their method name on `LlamaModel`~~ EDIT: I applied this rename. if we're breaking the API anyway, we might as well fix the name.

### DIFF
--- a/llama-cpp-2/src/lib.rs
+++ b/llama-cpp-2/src/lib.rs
@@ -82,12 +82,6 @@ pub enum ChatTemplateError {
     Utf8Error(#[from] std::str::Utf8Error),
 }
 
-enum InternalChatTemplateError {
-    Permanent(ChatTemplateError),
-    /// the buffer was too small.
-    RetryWithLargerBuffer(usize),
-}
-
 /// Failed to Load context
 #[derive(Debug, Eq, PartialEq, thiserror::Error)]
 pub enum LlamaContextLoadError {
--- a/llama-cpp-2/src/lib.rs
+++ b/llama-cpp-2/src/lib.rs
@@ -69,9 +69,14 @@ pub enum LLamaCppError {
 /// There was an error while getting the chat template from a model.
 #[derive(Debug, Eq, PartialEq, thiserror::Error)]
 pub enum ChatTemplateError {
-    /// gguf has no chat template
-    #[error("the model has no meta val - returned code {0}")]
-    MissingTemplate(i32),
+    /// gguf has no chat template (by that name)
+    #[error("chat template not found - returned null pointer")]
+    MissingTemplate,
+
+    /// chat template contained a null byte
+    #[error("null byte in string {0}")]
+    NullError(#[from] NulError),
+
     /// The chat template was not valid utf8.
     #[error(transparent)]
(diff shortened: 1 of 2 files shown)


AsbjornOlling:
Oh and a note on the documentation changes.

I removed this section from the docs, because I thought that it was rambling, difficult to understand, and mostly related to the implementation details of `apply_chat_template`, and not `get_chat_template`

```
    /// ... If this fails, you may either want to fail to chat or pick the
    /// specific shortcode that llama.cpp supports templates it has baked-in directly into its codebase
    /// as fallbacks when the model doesn't contain. NOTE: If you don't specify a chat template, then
    /// it uses chatml by default which is unlikely to actually be the correct template for your model
    /// and you'll get weird results back.
```

I also added a recommendation to use minijinja, to supplement the existing docs instructing people to use `apply_chat_template`. Most chat templates won't work at all or give very poor results with llama.cpp's hacky template rendering. 
Meanwhile, minijinja is a pure rust implementation of jinja template rendering, that was created and is maintained by the original creator of jinja, and is being used in production by [huggingface](https://github.com/huggingface/text-generation-inference/tree/0759ec495e15a865d2a59befc2b796b5acc09b50). Personally I think this is a better way to go :innocent:

MarcusDunn:
this looks great. Thanks for the PR!

MarcusDunn (approved):

AsbjornOlling:
Whoops... Looks like I accidentally made `chat_template` a private method when I renamed it  :sweat_smile: 

Is fixed now.

AsbjornOlling:
Found another (admittedly minor) error: some other methods' documentation comments still referred to `LlamaModel::get_chat_template`, which was replaced by `LlamaModel::chat_template` (for consistency with the naming in llama.h)

PR #706, opened 2025-04-10
Add llama-grammar.h for bindgen

yujonglee:
Having `llama-grammar` is super usful, for ex, validating GBNF (https://github.com/ggml-org/llama.cpp/blob/master/examples/gbnf-validator/gbnf-validator.cpp).


MarcusDunn:
Thanks for the PR! 

Unfortunately, this adds extra maintenance burden that I do not want this project to take on - keeping up to date with the "stable" llama.h is already a resonable amount of effort.

Issue #719, opened 2025-04-30
Support `qwen3` model architecture


vlad-ivanov-name:
This is supported in llama.cpp since https://github.com/ggml-org/llama.cpp/commit/d3bd7193ba66c15963fd1c59448f22019a8caf6e

agavrel:
`error loading model: error loading model architecture: unknown model architecture: 'qwen3moe'`

fellhorn:
I locally tried updating the `llama.cpp` submodule, see https://github.com/utilityai/llama-cpp-rs/pull/720.
Qwen3 then worked with CPU and CUDA builds.

AsbjornOlling:
@fellhorn Beat me to it, I guess. Can confirm it works on vulkan, too.

FWIW I just used the `update-llama-cpp-2025-04-29` branch on this repo to get a later llama.cpp version for Qwen3. They're automatically generated every day, so that's pretty convenient. But - I ran into CI issues because newer version of llama.cpp depend on libcurl.

MarcusDunn:
unable to publish after #721.
see: https://github.com/utilityai/llama-cpp-rs/actions/runs/14760441839

MarcusDunn:
thanks to @fellhorn and @AsbjornOlling, qwen 3 should be working!

PR #721, opened 2025-04-30
Bump llama.cpp, disable CURL

AsbjornOlling:
Sets LLAMA_CURL=OFF in build.rs.

Currently, llama-cpp-rs will fail to build on more recent commits of llama.cpp, if the libcurl libraries aren't installed, since LLAMA_CURL has changed defaults to being on by default.

If I am not mistaken, curl is only used in common.h, which is considered out of scope for this library, so we might as well disable it always.

This also bumps the version of llama.cpp to the most recent, which adds support for qwen3, closing #719

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -267,6 +267,7 @@ fn main() {
     config.define("LLAMA_BUILD_TESTS", "OFF");
     config.define("LLAMA_BUILD_EXAMPLES", "OFF");
     config.define("LLAMA_BUILD_SERVER", "OFF");
+    config.define("LLAMA_CURL", "OFF");
 
     config.define(
         "BUILD_SHARED_LIBS",


MarcusDunn (approved):

MarcusDunn:
seems like there's some build issues. Could you take a look? generally it's a missing file in the include.

AsbjornOlling:
Hopefully the dockerfile issue is fixed by #722

MarcusDunn:
Apologies, I wasn't clear. 

The issue is with publishing. You should be able to reproduce with `cargo publish --dry-run`

jiabochao:
It seems that the file minja/chat-template.hpp cannot be found.

```
CMake Error at common/CMakeLists.txt:55 (add_library):
    Cannot find source file:

      minja/chat-template.hpp

```
image

fellhorn:
I noticed that `common/minja/chat-template.hpp` exists in the llama.cpp repo and git submodule within `llama-cpp-sys-2` but is not there in the target build directory during `cargo publish`:

```bash
> ls ~/workspace/llama-cpp-rs/target/package/llama-cpp-sys-2-0.1.103/llama.cpp/common
arg.cpp            CMakeLists.txt              json-schema-to-grammar.h  sampling.h
arg.h              common.cpp                  llguidance.cpp            speculative.cpp
base64.hpp         common.h                    log.cpp                   speculative.h
build-info.cpp.in  console.cpp                 log.h                     stb_image.h
chat.cpp           console.h                   ngram-cache.cpp
chat.h             json.hpp                    ngram-cache.h
cmake              json-schema-to-grammar.cpp  sampling.cpp
```

How is the common folder copied there?

PR #757, opened 2025-06-18
fix(build): update CUDA linking strategy for Windows.

DAIMENGEN:
This build script is modified to properly handle platform-specific CUDA linking:

On Windows:

NVIDIA’s CUDA SDK does not ship culibos.lib, and static libraries like cublas_static.lib are typically unavailable.

Attempting to statically link cudart_static on Windows will fail due to missing symbols provided by culibos.

Therefore, on Windows, dynamic linking is preferred. The .lib files (like cudart.lib) act as import libraries for corresponding .dll files (e.g., cudart64_*.dll).

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -410,28 +410,44 @@ fn main() {
     println!("cargo:rustc-link-search={}", build_dir.display());
 
     if cfg!(feature = "cuda") && !build_shared_libs {
+        // Re-run build script if CUDA_PATH environment variable changes
         println!("cargo:rerun-if-env-changed=CUDA_PATH");
 
+        // Add CUDA library directories to the linker search path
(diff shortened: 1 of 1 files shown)


MarcusDunn:
I'd like to see it rebased on top of main because there was a resonably sized version bump of upstream that just landed. Otherwise this looks good to me

DAIMENGEN:
Thanks for the feedback!

I'll rebase the PR on top of the latest main shortly to incorporate the recent upstream changes.

Regarding the Windows check, I'll look into getting it to pass as well. It makes total sense to have a reliable CI signal for Windows going forward — totally agree that it will help catch regressions early. I'll update here once I’ve made progress on that.

Thanks again for reviewing and for maintaining the project!

DAIMENGEN:
I will close the current PR for now and open a new one later.

PR #758, opened 2025-06-19
Make LlamaSampler::grammar() return an Option

MatthewCash:
Currently, any calls to `llama_sampler_init_grammar()` or `llama_sampler_init_grammar_lazy()` are always expected to return a valid pointer to a sampler. This means that if an invalid grammar is provided, both of those functions would return `nullptr` and the program will segfault after attempting to dereference it. 

This change makes the grammar sampler functions return an `Option<LlamaSampler>` so that a failure to create a grammar can be handled.


MarcusDunn (approved):
great change!

PR #763, opened 2025-07-06
Add swa_full

yujonglee:
https://github.com/ggml-org/llama.cpp/blob/6491d6e4f1caf0ad2221865b4249ae6938a6308c/common/common.h#L332

https://github.com/ggml-org/llama.cpp/blob/6491d6e4f1caf0ad2221865b4249ae6938a6308c/src/llama-context.cpp#L2159


MarcusDunn:
thanks for the PR!

PR #774, opened 2025-07-21
Support llguidance

### LINKED ISSUE #701: LLGuidance support with feature flag
https://github.com/ggml-org/llama.cpp/blob/master/docs/llguidance.md

> enable LLGuidance support, build llama.cpp with the LLAMA_LLGUIDANCE option:
```
cmake -B build -DLLAMA_LLGUIDANCE=ON
make -C build -j
```

This would be great addition.

yujonglee:
Resolves #701

### DIFF
--- a/llama-cpp-2/Cargo.toml
+++ b/llama-cpp-2/Cargo.toml
@@ -30,6 +30,7 @@ openmp = ["llama-cpp-sys-2/openmp"]
 sampler = []
 # Only has an impact on Android.
 android-shared-stdcxx = ["llama-cpp-sys-2/shared-stdcxx"]
+llguidance = ["llama-cpp-sys-2/llguidance"]
 
 
 [target.'cfg(all(target_os = "macos", any(target_arch = "aarch64", target_arch = "arm64")))'.dependencies]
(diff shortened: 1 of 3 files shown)


MarcusDunn (commented):
Looking into this, I can see why this would be a very cool feature. 

However, I don't think this approach will do what you're expecting. 

The only place the llguidance API's are exposed is in `common/sampling.h`, which we do not generate bindings to. I do not want this to change as keeping up to date with the main API is difficult enough as is with as little maintenance effort as this project currently receives.

On the bright side, llguidance is a rust library and I think a llama-cpp-rs + llguidance library would be pretty easy to build and maintain. The whole concept of building a rust library (llguidance) as part of a c++ library (llama.cpp) to be used by a rust library (llama-cpp-rs) strikes me as the wrong approach. 

Unless I'm mistaken I will probably not merge this PR.

yujonglee:
Agreed. Thanks for the input.

fellhorn:
Churning in from the sidelines: We internally developed a test runtime based upon llama-cpp-rs and also support sampling with llguidance. We do all the (constrained) sampling purely on the Rust side. I also don't see a need to expose this through llama-cpp-rs - it works pretty well this way.

Let me know if you have any questions!

yujonglee:
@fellhorn 

Nice. My llama-cpp-rs inference code is part of open-source project:
https://github.com/fastrepl/hyprnote/blob/main/crates/llama/src/lib.rs

Maybe we can share notes. Also if you kindly help us implement it, that would be super awesome. :)

Issue #784, opened 2025-07-27
OpenBSD Support


donpellegrino:
The build.rs conditionals do not account for OpenBSD. Tested against OpenBSD 7.7. The seems to be easily fixed for Vulkan support but OpenMP support might take more work. Documentation for the fix should also list which packages needed to be added to the base system to fulfill dependencies.

Issue #725, opened 2025-05-02
Plans to support AMD


ljt019:
Are there any plans to support inference with AMD gpus? As far as I can tell currently both here and in Candle there is only support for NVIDIA gpus.

The closest repo I found working on support/bindings for rust was this crate here:
https://github.com/radudiaconu0/rocm-rs

MarcusDunn:
This should be an upstream issue. If llama.cpp has support for AMD, we can realistically support it - PRs are always welcome.

PR #785, opened 2025-07-27
Add rocm hip support

donpellegrino:
#725 & #784 could be addressed by these changes. Note that the change was tested against an upstream llama.cpp, so it may be best to wait on testing until llama.cpp is updated as part of the normal flow. Testing was done on AMD Developer Cloud with Ubuntu 24.04.1 LTS, AMD ROCm 6.4.1, and an AMD MI300X. This is just a proof-of-concept tested in the one environment. It needs review and execution within a test suite. Let me know if there is a test suite I should run it through.


MarcusDunn:
https://github.com/utilityai/llama-cpp-rs/actions/runs/16555633011/job/46873897581?pr=785 should be passing. Looks like formatting is the first step.

PR #786, opened 2025-07-27
Bump llama.cpp to b6002

fellhorn:
Bump llama.cpp to [b6002](https://github.com/ggml-org/llama.cpp/commit/89d1029559bd2968f76db854f9f113d73e34527c).

### DIFF
--- a/llama-cpp-2/src/timing.rs
+++ b/llama-cpp-2/src/timing.rs
@@ -36,6 +36,7 @@ impl LlamaTimings {
                 t_eval_ms,
                 n_p_eval,
                 n_eval,
+                n_reused,
             },
         }
     }


fellhorn:
A prerequisite for multimodal support - a couple of bug fixes have been added in the meantime.

MarcusDunn:
merging main may fix.

fellhorn:
There is a chicken-egg issue in CI: `cargo publish --dry-run --package llama-cpp-2` requires that the new version of `llama-cpp-sys-2` has already been published. The step fails at it uses the currently published version instead of the workspace version.

How was that solved before?

I also noticed that `llama-cpp-sys-2` is not pinned as a dependency when publishing the crate. Maybe `cargo-release` could simplify that process.

fellhorn:
The other failure seems to be a temporary (?) issue when pulling `nvcr.io/nvidia/cuda:12.3.1-devel-ubuntu22.04`

PR #789, opened 2025-07-31
Fix panic introduced upstream if debug log level is disabled

### LINKED ISSUE #788: Panic when debug log level is disabled
https://github.com/utilityai/llama-cpp-rs/pull/760 introduces a panic because the previous log level is never actually initialized and then we receive a CONT which tries to resume LOG_LEVEL_NONE.

The problem is in lib.rs:
```
    if !log_state.is_enabled_for_level(level) {
        return;
    }
```

Before you early return, you still have to record the previous log level and change `is_enabled_for_level` to not just blindly return True but actually use the previous level.

vlovich:
If the dispatcher has disabled a log level, a CONT for that log level causes a panic.

Fixes #788

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -670,6 +670,7 @@ dependencies = [
  "thiserror",
  "tracing",
  "tracing-core",
+ "tracing-subscriber",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


vlovich:
@MarcusDunn I believe the failing checks is due to bitrot & unrelated to my change?

MarcusDunn (approved):

PR #790, opened 2025-08-01
Multimodality support

### LINKED ISSUE #744: Multimodal support (`libmtmd`)
llama.cpp [recently refactored](https://github.com/ggml-org/llama.cpp/pull/13311) their [support for multimodal input](https://github.com/ggml-org/llama.cpp/tree/master/tools/mtmd#multimodal-support-in-llamacpp) and it's getting closer to being stabilized, especially as there is  now support in the server.

It would be great adding Rust bindings for `/tools/mtmd` to allow using multimodal input also through llama-cpp-rs.

fellhorn:
Add bindings for the libmtmd multimodality support of llama-cpp. This allows to use models like e.g. Gemma3 with image or audio input. The bindings are only created with a new `mtmd` feature. To explain the usage, I added a simple example loosely based on the `mtmd-cli.cpp`.

To run the mtmd example, you first need to download the model gguf file and the multimodal projection file, e.g. for Gemma3 you may use:

```sh
wget https://huggingface.co/unsloth/gemma-3-4b-it-GGUF/resolve/main/gemma-3-4b-it-Q4_K_M.gguf \
https://huggingface.co/unsloth/gemma-3-4b-it-GGUF/resolve/main/mmproj-F16.gguf
```

To then run the example on CPU, provide an image file `my_image.jpg` and run:

```sh
cargo run --release --example mtmd -- \
  --model ./gemma-3-4b-it-Q4_K_M.gguf \
  --mmproj ./mmproj-F16.gguf \
  --image my_image.jpg \
  --prompt "What is in the picture?" \
  --no-gpu \
  --no-mmproj-offload \
  --marker "<start_of_image>"
```

Closes #744

### DIFF
--- a/llama-cpp-sys-2/Cargo.toml
+++ b/llama-cpp-sys-2/Cargo.toml
@@ -9,6 +9,7 @@ links = "llama"
 
 include = [
     "wrapper.h",
+    "wrapper_mtmd.h",
     "build.rs",
     "/src",
 
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -277,6 +277,7 @@ fn main() {
     // Configure mtmd feature if enabled
     if cfg!(feature = "mtmd") {
         bindings_builder = bindings_builder
+            .header("wrapper_mtmd.h")
             .allowlist_function("mtmd_.*")
             .allowlist_type("mtmd_.*");
     }
@@ -432,6 +433,7 @@ fn main() {
         .expect("Failed to write bindings");
 
     println!("cargo:rerun-if-changed=wrapper.h");
+    println!("cargo:rerun-if-changed=wrapper_mtmd.h");
 
     debug_log!("Bindings Created");
 
--- a/llama-cpp-sys-2/wrapper.h
+++ b/llama-cpp-sys-2/wrapper.h
@@ -1,3 +1,1 @@
 #include "llama.cpp/include/llama.h"
-#include "llama.cpp/tools/mtmd/mtmd.h"
-#include "llama.cpp/tools/mtmd/mtmd-helper.h"
--- a/llama-cpp-sys-2/wrapper_mtmd.h
+++ b/llama-cpp-sys-2/wrapper_mtmd.h
@@ -1,0 +1,2 @@
+#include "llama.cpp/tools/mtmd/mtmd.h"
+#include "llama.cpp/tools/mtmd/mtmd-helper.h"


llama-cpp-2/src/mtmd.rs
```diff
@@ -0,0 +1,899 @@
...
+
+impl Default for MtmdContextParams {
+    fn default() -> Self {
+        Self {
+            use_gpu: false,
+            print_timings: true,
+            n_threads: 4,
+            media_marker: CString::new(mtmd_default_marker()).unwrap_or_default(),
+        }
+    }
```
MarcusDunn, line 86: this seems like it should rely on `mtmd_context_params_default`

llama-cpp-2/src/mtmd.rs
```diff
@@ -0,0 +1,899 @@
...
+            llama_cpp_sys_2::mtmd_init_from_file(
+                path_cstr.as_ptr(),
+                text_model.model.as_ptr(),
+                ctx_params,
+            )
+        };
+
+        if context.is_null() {
+            return Err(MtmdInitError::NullResult);
+        }
```
MarcusDunn, line 170: redundant check.

llama-cpp-2/src/mtmd.rs
```diff
@@ -0,0 +1,899 @@
...
+
+        let context = NonNull::new(context).ok_or(MtmdInitError::NullResult)?;
+        Ok(Self { context })
+    }
+
+    /// Check whether non-causal attention mask is needed before `llama_decode`.
+    #[must_use] pub fn decode_use_non_causal(&self) -> bool {
+        unsafe { llama_cpp_sys_2::mtmd_decode_use_non_causal(self.context.as_ptr()) }
+    }
```
MarcusDunn, line 179: a safety comment would be great. (self.context is not null should be sufficent?)

llama-cpp-2/src/mtmd.rs
```diff
@@ -0,0 +1,899 @@
...
+        if result == 0 {
+            Ok(())
+        } else {
+            Err(MtmdEncodeError::EncodeFailure(result))
+        }
+    }
+}
+
+unsafe impl Send for MtmdContext {}
+unsafe impl Sync for MtmdContext {}
```
MarcusDunn, line 308: These impls need some sort of comment on why this is safe.

llama-cpp-2/src/mtmd.rs
```diff
@@ -0,0 +1,899 @@
...
+            return None;
+        }
+
+        let chunk_ptr =
+            unsafe { llama_cpp_sys_2::mtmd_input_chunks_get(self.chunks.as_ptr(), index) };
+
+        if chunk_ptr.is_null() {
+            None
+        } else {
+            // Note: We don't own this chunk, it's owned by the chunks collection
```
MarcusDunn, line 611: should this return a reference then?

llama-cpp-2/src/mtmd.rs
fellhorn: Removed in [d025465](https://github.com/utilityai/llama-cpp-rs/pull/790/commits/d0254656235576803f9e41067d180381e37c89c5)

llama-cpp-2/src/mtmd.rs
```diff
@@ -0,0 +1,899 @@
...
+    /// based on a hash of the bitmap data.
+    #[must_use] pub fn id(&self) -> Option<String> {
+        let ptr = unsafe { llama_cpp_sys_2::mtmd_bitmap_get_id(self.bitmap.as_ptr()) };
+        if ptr.is_null() {
+            None
+        } else {
+            unsafe { CStr::from_ptr(ptr) }
+                .to_string_lossy()
+                .into_owned()
+                .into()
```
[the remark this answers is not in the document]
fellhorn: Cleaned it up in d025465

llama-cpp-2/src/mtmd.rs
fellhorn: Good catch, I actually meant `MtmdContextParams` and that is already `Send + Sync`.  Afais `mtmd_context` is not thread safe. Removed in d025465

fellhorn:
@MarcusDunn, thanks for the thorough review. I think I addressed all your comments in d025465 and f149f11.
Especially your comment regarding feature gating everything related to mtmd should be addressed with f149f11:
I added another `wrapper_mtmd.h`, which will only be used when building with the mtmd feature.

Please let me know if there are still mtmd dependencies that I might have missed. I assume that the `include` in the Cargo.toml should be fine?

llama-cpp-2/src/mtmd.rs
```diff
@@ -0,0 +1,931 @@
...
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MtmdInputChunkType {
+    /// Text input chunk
+    Text = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_TEXT as isize,
+    /// Image input chunk
+    Image = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_IMAGE as isize,
+    /// Audio input chunk
+    Audio = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_AUDIO as isize,
+}
```
MarcusDunn, line 38: is the cast required?

MarcusDunn (approved):
LGTM. Have the non-gemma models been tested? (I will merge regardless as we can always fix that later).

Feel free to address the comments, I'll merge a couple days from now to see if others want to give this branch some testing before I merge.

haixuanTao:
Can confirm the model mention above is now working wonders!

Thanks a lot!

haixuanTao:
More of an optimisation thing, but are we using the maximum number of batches for encoding images and audio? 

```rust
        self.n_past = chunks.eval_chunks(&self.mtmd_ctx, context, 0, 0, 1, true)?; // batch size of image encoding
```

It seems that this could be higher than 1, but not an expert on this

haixuanTao:
FYI, I tried to pass image+audio at the same time and it didn't work as there were only one marker within the prompt :D 

Not an urgent bug

haixuanTao:
> More of an optimisation thing, but are we using the maximum number of batches for encoding/decoding images and audio?
> 
> ```rust
>         self.n_past = chunks.eval_chunks(&self.mtmd_ctx, context, 0, 0, 1, true)?; // batch size of image decoding
> ```
> 
> It seems that this could be higher than 1, but not an expert on this
> 
> Related: [ggml-org/llama.cpp#14527](https://github.com/ggml-org/llama.cpp/issues/14527) if you search through the issue he's actually decoding at 64 n_tokens_batch

I am able to batch by 10 using qwen-omni, reducing the decoding time from 120ms-> 50ms, but I truly have no idea why most of time when giving a `n_batch` a random batch size it gives me the following error:

```bash
image slice encoded in 326 ms
decoding image batch 1/3, n_tokens_batch = 20
image decoded (batch 1/3) in 52 ms
decoding image batch 2/3, n_tokens_batch = 20
image decoded (batch 2/3) in 53 ms
decoding image batch 3/3, n_tokens_batch = 20
image decoded (batch 3/3) in 53 ms
get_logits_ith: invalid logits id 0, reason: batch.logits[0] != true
```

haixuanTao:
It's super fast in cuda. But it's a llama.cpp thing apparently

haixuanTao:
Related: https://github.com/ggml-org/llama.cpp/issues/15426

altunenes:
> Related: [ggml-org/llama.cpp#15426](https://github.com/ggml-org/llama.cpp/issues/15426)

 thanks. experiencing exactly the same issue. tried many things. I thought it was my fault. I've been working on it since yesterday. I tried running it on different threads, but no matter what I did, my optimization only improved it by 2-3 seconds at most. Also, I was doing all the image/video processing on the GPU side (using wgpu). 
Not tried Vulkan though

PR #798, opened 2025-08-07
Extend embedding examples to Qwen, miniLM, GTE, E5, JINA, Nomic

srv1n:
This crate can support multiple embedding model architectures. This PR adds reference examples to these models Qwen, miniLM, GTE, E5, JINA, Nomic. There are also python scripts that verify the embeddings generated are close to the official HuggingFace scripts usually offered by the model authors. 

Slight discrepancies may occur due to the quantization of the GGUF models, but the results should be very close to the official implementations based on a cosine similarity check. 

Tests are run on first paragraph of top 50(2025) wikipedia articles, made avaialble in the `wikipedia_embedding_dataset.csv` file. Once tests are run the top 5 texts are saved to a CSV file for dimension-by-dimension comparison. The results are also saved in a markdown file with the full report of the benchmark. A sample from the real run is here [results/benchmark_report_20250806_230730.md](results/benchmark_report_20250806_230730.md).

There is a detailed Readme file inside that walks through methods, scripts (both to download and run an automated report generation comparing speeds and similarity)

### Cosine Similarity vs Python Baseline

| Model | Rust Similarity |
|-------|----------------|
| BGE-Small | 0.9999 |
| E5-Base | 1.0000 |
| GTE-Base | 0.9999 |
| JINA-v2-Base | 0.9999 |
| Qwen3 | 0.9997 |
| Nomic-v1.5 | 0.9800 |
| MxBAI-Large | 0.9412 |


MarcusDunn:
My instinct is that this adds more code than its worth to the repo - beyond better testing (which is a good goal) is there other reasons to add this?

srv1n:
True. I thought it may be helpful to folks using this repo to have reference code for other embedding implementation other than bge.

If it doesnt make sense, happy to just add reference snippets here and we can close out the issue.

MarcusDunn:
I think adding some snippets to the module documentation may be a better way to go.

PR #819, opened 2025-09-01
Multimodality improvements

fellhorn:
Improves a few issues in the mtmd example and wrappers that remained from #790 :

* fixes the batch size param (thanks @haixuanTao for spotting this)
* addresses remaining nits from @MarcusDunn 
* clippy cleanup

### DIFF
--- a/examples/mtmd/src/mtmd.rs
+++ b/examples/mtmd/src/mtmd.rs
@@ -51,7 +51,7 @@ pub struct MtmdCliParams {
     #[arg(short = 't', long = "threads", value_name = "N", default_value = "4")]
     pub n_threads: i32,
     /// Number of tokens to process in a batch during eval chunks
-    #[arg(long = "batch-size", value_name = "b", default_value = "64")]
+    #[arg(long = "batch-size", value_name = "b", default_value = "1")]
     pub batch_size: i32,
     /// Maximum number of tokens in context
     #[arg(long = "n-tokens", value_name = "N", default_value = "4096")]
--- a/llama-cpp-2/src/mtmd.rs
+++ b/llama-cpp-2/src/mtmd.rs
@@ -25,18 +25,18 @@ use crate::token::LlamaToken;
 /// let audio_chunk = MtmdInputChunkType::Audio;
 ///
 /// assert_eq!(text_chunk, MtmdInputChunkType::Text);
-/// assert_eq!(text_chunk as u32, llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_TEXT);
+/// assert_eq!(text_chunk, llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_TEXT.into());
 /// assert_ne!(text_chunk, image_chunk);
 /// ```
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(u32)]
 pub enum MtmdInputChunkType {
     /// Text input chunk
-    Text = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_TEXT,
+    Text = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_TEXT as _,
     /// Image input chunk
-    Image = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_IMAGE,
+    Image = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_IMAGE as _,
     /// Audio input chunk
-    Audio = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_AUDIO,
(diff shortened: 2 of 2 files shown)


fellhorn:
> I mean I have tried multiple batch size, but it seems that the actual batch size that works for images media or audio media can sometimes make the predictions not work. 64 typically does not work for me. 10 work but only on images with Owen2.5-omni.
> 
> I wonder if it's something on our end or something different.

Do you see an error or are the results just wrong? I tried Gemma-3-4b with this branch and had positive results with a BS of 1, 16 and 256. The only thing I noticed is, that the batch size needs to be a divisor of the number of visual tokens per image. For Gemma that's 256.

If one does not pick a divisor, decoding fails with

```
init: the tokens of sequence 0 in the input batch have inconsistent sequence positions:
- the last position stored in the memory module of the context (i.e. the KV cache) for sequence 0 is X = 260
- the tokens for sequence 0 in the input batch have a starting position of Y = 267
it is required that the sequence positions remain consecutive: Y = X + 1
decode: failed to initialize batch
llama_decode: failed to decode, ret = -1
failed to decode text
failed to eval chunk 2
Error: EvalFailure(-1)
```

llama-cpp-2/src/mtmd.rs
```diff
@@ -25,16 +25,18 @@ use crate::token::LlamaToken;
...
 pub enum MtmdInputChunkType {
     /// Text input chunk
-    Text = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_TEXT as isize,
+    Text = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_TEXT,
     /// Image input chunk
-    Image = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_IMAGE as isize,
+    Image = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_IMAGE,
     /// Audio input chunk
-    Audio = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_AUDIO as isize,
+    Audio = llama_cpp_sys_2::MTMD_INPUT_CHUNK_TYPE_AUDIO,
```
fellhorn, line 39: Just saw the Github Action run - this does not work on Windows due to different bindgen behavior.

I'll take another look tomorrow or revert to `as isize`

MarcusDunn: you can get around this with `as _`, if there's a corresponding bindgen type, I would prefer that.

examples/mtmd/src/mtmd.rs
```diff
@@ -50,6 +50,9 @@ pub struct MtmdCliParams {
     /// Number of threads
     #[arg(short = 't', long = "threads", value_name = "N", default_value = "4")]
     pub n_threads: i32,
+    /// Number of tokens to process in a batch during eval chunks
+    #[arg(long = "batch-size", value_name = "b", default_value = "64")]
```
haixuanTao, line 54: Can we keep the default batch size to 1 ?

fellhorn:
> Basically it can either follow the prompt or the audio. And so, if the audio is like how many duck is there in the image it should work.
> 
> But it seems that the audio overwrite the prompt

I also tried a couple of prompts, where audio and image were not related. It still worked fairly well and described them separately from each other. What I noticed though: Qwen2.5-Omni seems rather weak at answering complex questions about the audio. Most likely because the vision and audio encoder only provide a compressed view to the model.

@haixuanTao Have you tried your prompts with the Qwen demo on the [hf website](https://huggingface.co/spaces/Qwen/Qwen2.5-Omni-7B-Demo) or with vllm? It would be interesting to see if it's an issue with the model or one with llama.cpp or the Rust bindings.

fellhorn:
> @fellhorn Would you like this to be merged as is? I have no issues on my end but it seems like there's still active discussion.

@MarcusDunn I think we can go ahead with the PR and merge it. The discussion is not about the enhancements here. If some new changes come out of it, I'll create a separate PR.

haixuanTao:
> > Basically it can either follow the prompt or the audio. And so, if the audio is like how many duck is there in the image it should work.
> > But it seems that the audio overwrite the prompt
> 
> I also tried a couple of prompts, where audio and image were not related. It still worked fairly well and described them separately from each other. What I noticed though: Qwen2.5-Omni seems rather weak at answering complex questions about the audio. Most likely because the vision and audio encoder only provide a compressed view to the model.
> 
> @haixuanTao Have you tried your prompts with the Qwen demo on the [hf website](https://huggingface.co/spaces/Qwen/Qwen2.5-Omni-7B-Demo) or with vllm? It would be interesting to see if it's an issue with the model or one with llama.cpp or the Rust bindings.

I have only tried simple audio translation, as of now, and image grounding. As it's not an instruct model, it's not going to be very good at conversation I think.

I mean in general it behaves as expected. Except for the batch size which I feel might be more than what other ppl are doing.

PR #820, opened 2025-09-04
Expose seperator token

emilnorsker:
The separator token is required in BERT style rerankers i.e.: `"<cls><query><sep><document><sep>"`

Normally cls token is also required for rerankers but it is currently deprecated in llama.cpp with a [comment](https://github.com/ggml-org/llama.cpp/blob/master/include/llama.h#L992) stating that it is the same as the bos token for a model, and thus not necessary.


MarcusDunn:
publishing now! Thanks for the PR

PR #833, opened 2025-09-21
feat(opencl): add optional OpenCL backend support via Cargo feature

admiralakber:
Hey mate,

Simple yet high value add, I thought I should upstream it. I have tested this on Linux and Android. The OpenCL backend is much more reliable than Vulkan.

## ✨ Summary

- Introduce an `opencl` feature for both crates, package ggml-opencl sources, and enable GGML_OPENCL in build.rs with per-OS linking to the OpenCL loader. Keeps changes consistent with existing CUDA/Vulkan patterns.

## 🔎 Diagnosis

- Place behind rust feature flag to avoid breaking existing builds
- Grab defaults from llama.cpp upstream: https://github.com/ggml-org/llama.cpp/blob/master/docs/backend/OPENCL.md

## 🚀 Changes

- New feature flag on `llama-cpp-2/Cargo.toml` and added new globbing to `llama-cpp-sys-2/Cargo.toml`
- Little tweak to the llama-cpp-sys-2 `README.md` should help SEO if that matters anymore
- Most of the work in `llama-cpp-2-sys/build.rs` - matched the existing patterns

## ☁️ What you’ll need to do

- Just build with the `opencl` feature flag

## 🧪 Smoke test

```
cargo run -p llama-cpp-2 --example usage --features opencl -- ${PATH_TO_MODEL} --verbose 
```

## ✅ Acceptance

- If someone else can do a manual test that would be great. A Linux box with a dediated GPU would be strong validation. 
  - Note: It might be that the llama.cpp OpenCL backend is only supported for Android which I can report is working great.
- A Windows or Mac build just to confirm the rust link commands in `build.rs` are correct.

### DIFF
--- a/llama-cpp-2/Cargo.toml
+++ b/llama-cpp-2/Cargo.toml
@@ -32,6 +32,7 @@ sampler = []
 # Only has an impact on Android.
 android-shared-stdcxx = ["llama-cpp-sys-2/shared-stdcxx"]
 mtmd = ["llama-cpp-sys-2/mtmd"]
+opencl = ["llama-cpp-sys-2/opencl"]
 
 
 [target.'cfg(all(target_os = "macos", any(target_arch = "aarch64", target_arch = "arm64")))'.dependencies]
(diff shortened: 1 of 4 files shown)


admiralakber:
Hey, looks like the build failure on linux/amd64 is that the runner ran out of space?

Issue #826, opened 2025-09-09
Allow decoding Control Tokens to String


ju6ge:
Currently it is not possible to decode control code tokens:

> https://github.com/utilityai/llama-cpp-rs/blob/1ff8410d2609dce9cdb50aa81b9d0d3dcc05f289/llama-cpp-2/src/model.rs#L382-L436)

The reason being that `token_to_bytes_with_size` returns an empty bytes vector when trying to get the corresponding string for an token like `model.token_eos()` … if I comment it out I can decode the token just fine. Is there a particular reason for this behavior? The comment in the code is a bit unclear why this is handled this way.

The reason this would be useful is that if I want to generate a grammar for the response of model i need to include the `eos` token in order to terminate the generation gracefully. But different models use different `eos` tokens so without knowing what string the eos token corresponds to I am unable to generate the correct grammar.

ju6ge:
I think this issues was address and can be closed 🥳

PR #856, opened 2025-11-03
Add gate to disable control token decoding

ju6ge:
This PR implements the required changes to address #826

For easy opt in to the original behavior I added a check to an environment variable. This would make the behavior controllable 
at runtime. I can easily change this to be a compile time opt in by using a rust feature flag. Let me know what you think …

Kind regards
ju6ge

### DIFF
--- a/llama-cpp-2/src/model.rs
+++ b/llama-cpp-2/src/model.rs
@@ -115,6 +115,10 @@ pub enum AddBos {
 pub enum Special {
     /// Allow tokenizing special and/or control tokens which otherwise are not exposed and treated as plaintext. Does not insert a leading space.
     Tokenize,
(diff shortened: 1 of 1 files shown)


MarcusDunn (changes requested):
Thanks for the PR. See comment.

MarcusDunn (approved):

MarcusDunn (changes requested):
I think we should deprecate the current function (as well as all of our functions that call it), create a new versions that call `token_to_peice` without any special logic. (no messing around with attrs, as I imagine this is pretty application specific)

This aligns better with the goal of being a safe wrapper and once the deprecated functions are removed there is less code to maintain.

`Special` should remain an enum of two variants.

PR #861, opened 2025-11-11
Add ability to select backend devices

kusaanko:
Now, LlamaModelParams has a field `main_gpu` but it is not utilized because we can't set `split_mode` to `LLAMA_SPLIT_MODE_NONE`.

This PR  makes `split_mode` settable.
Moreover, this PR allows you to specify which backend devices to use in safe Rust.

The GGML backend device indices are not modified until you unload ggml backends, so the indices are used to specify devices.

### DIFF
--- a/llama-cpp-2/src/model/params.rs
+++ b/llama-cpp-2/src/model/params.rs
@@ -411,7 +411,7 @@ impl LlamaModelParams {
 /// assert_eq!(params.vocab_only(), false, "vocab_only should be false");
 /// assert_eq!(params.use_mmap(), true, "use_mmap should be true");
 /// assert_eq!(params.use_mlock(), false, "use_mlock should be false");
-/// assert_eq!(params.split_mode(), LlamaSplitMode::Layer, "split_mode should be LAYER");
+/// assert_eq!(params.split_mode(), Ok(LlamaSplitMode::Layer), "split_mode should be LAYER");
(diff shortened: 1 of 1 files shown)


llama-cpp-2/src/model/params.rs
```diff
@@ -22,15 +22,23 @@ pub enum LlamaSplitMode {
...
-impl From<i32> for LlamaSplitMode {
-    fn from(value: i32) -> Self {
+/// An error that occurs when unknown split mode is encountered.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct LlamaSplitModeParseError(i32);
+
+/// Create a `LlamaSplitMode` from a `c_int`.
+///
+/// # Errors
+/// Returns `()` if the value does not correspond to a valid `LlamaSplitMode`.
```
MarcusDunn:
```suggestion
/// Returns `LlamaSplitModeParseError` if the value does not correspond to a valid `LlamaSplitMode`.
```

MarcusDunn (approved):
left a suggestion to clean up docs. Rest looks good.

llama-cpp-2/src/model/params.rs
kusaanko: Done

MarcusDunn:
you can use `llama_cpp_sys_2::LLAMA_SPLIT_MODE_NONE as _` to cast to the correct platform type.

PR #868, opened 2025-11-21
add close_ggml_native feature

cyberdoors:
Adding the feature to disable GGML_NATIVE and enabling the close_ggml_native feature may result in a performance decrease, but it allows the executable to run on other computers.


AsbjornOlling:
A different, but related task is to support the `GGML_CPU_ALL_VARIANTS` and `GGML_BACKEND_DL` configuration variables. Setting those makes llama.cpp compile dynamic libraries for multiple cpu variants, and then allows selecting the appropriate one at runtime.

This is the best of both worlds: maximum portability and performance.

PR #874, opened 2025-11-28
Improve the grammar error handling

marek-hradil:
Grammar constructors have now pretty large number of exceptions on which they panic. Some of them are even not documented in the doc string of the functions, maybe because even llama.cpp does not mention them.

For example, when the `grammar_str` does not include `grammar_root`, upon calling `grammar_lazy` it just throws
```
Caused by:
  process didn't exit successfully: `...` (signal: 6, SIGABRT: process abort signal)
```
which is terrible for figuring out what is wrong (ive been there).

This PR fixes this, adding `GrammarError` and sanitizing the inputs to `grammar` and `grammar_lazy` functions.
The change is however *breaking* as we need a way to propagate the errors - the output type thus moved from `Option<LlamaSampler>` to `Result<LlamaSampler, GrammarError>`.

### DIFF
--- a/llama-cpp-2/src/lib.rs
+++ b/llama-cpp-2/src/lib.rs
@@ -156,6 +156,23 @@ pub enum EmbeddingsError {
     NonePoolType,
 }
 
+/// Errors that can occur when initializing a grammar sampler
+#[derive(Debug, Eq, PartialEq, thiserror::Error)]
+pub enum GrammarError {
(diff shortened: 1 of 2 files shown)


MarcusDunn:
the grammar code was created *just* as llama.cpp started it and has been untouched since. Thanks for taking the time to improve it!

PR #882, opened 2025-12-02
Use cargo's target-cpu configuration

AsbjornOlling:
Let's use the standard rust configuration stuff to control what cpu architectures to compile for.
E.g. setting `target-cpu=znver4` (for the ryzen chip in my laptop), or `target-cpu=native` (to compile for whatever the compiling machine's cpu is).

It seems like the behavior prior to this PR was:
- the `native` crate feature does nothing(??), except for an edgecase on aarch64 linux
- rust's `target-cpu` config is used in rustc as usual
- `GGML_NATIVE` is set to the default value selected by llama.cpp (usually `ON`, but `OFF` when cross-compiling or when compiling a reproducible build)

That resulted in simple compilations of llama-cpp-rs which contain pure-rust code built for baseline x86-64 with no fancy extensions, and the bundled C++ GGML code built for whatever extensions are available on the compiling machine.

This PR does the following:
- remove the `native` feature completely
- look through `CARGO_ENCODED_RUSTFLAGS` to pick out the value of `target-cpu`
- if `target-cpu` is `"native"`, enable `GGML_NATIVE`
- if `target-cpu` is anything else: 
  - get a cpu feature list from `CARGO_CFG_TARGET_FEATURE`
  - for each cpu feature, set the corresponding `GGML` config
  - set `CFLAGS` and `CXXFLAGS` to `-march=<target-cpu>` 

After this PR, both the pure rust code and the ggml-cpu code will be built with the same `target-cpu` configuration. The default `target-cpu` configuration is a very boring/conservative `x86-64`. This means that straightforward builds which use no additional configuration *used to* be built with `GGML_NATIVE`, and will after this PR be built without it.

**Builds that use CPU inference and don't specify a `target-cpu` will likely become slower**.

Hence, this is a breaking change, and it could be discussed whether it's a regression. I believe that the behavior implemented in this PR is correct.

This PR supersedes #868 

~~This PR is WIP, since the resulting binaries seem to still contain `AVX2` instructions even when I'm compiling with `RUSTFLAGS="-C target-cpu=x86-64"`. Not sure what's going on, but I'm learning and investigating...~~ solved. the remaining AVX2 instructions turned out to be directly included assembly.

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -514,7 +514,7 @@ fn main() {
         .and_then(|rustflags| {
             rustflags
                 .split('\x1f')
-                .find(|f| f.contains("target-cpu=") && !f.contains("target-cpu=native"))
+                .find(|f| f.contains("target-cpu="))
                 .and_then(|f| f.split("target-cpu=").nth(1))
                 .map(|s| s.to_string())
         });


MarcusDunn (approved):
This is fantastic. I'm happy to have the default case regress in exchange for this matching cargo conventions.

AsbjornOlling:
I have a few tiny code-cleanlyness changes that I'm going to push in a sec, so you might want to hold off on merging.

AsbjornOlling:
Spotted another bug, so I marked this PR as a draft again. I'll ping in here when I think it's ready :)

PR #980, opened 2026-04-02
fix(build): broken shared library symlinks on Linux with dynamic-link

ryaningli:
On Linux, CMake produces a soname symlink chain for shared libraries:
  libllama.so -> libllama.so.0 -> libllama.so.0.0.8233

The previous code used glob("*.so") which only matched the top-level symlinks (libllama.so), and std::fs::hard_link on a symlink creates another symlink rather than a hard link to the real file. This left target/release/ full of broken symlinks with no actual .so files.

Fix by:
- Changing glob pattern from "*.so" to "*.so*" to match all files in the soname chain
- Introducing copy_lib_asset() which uses symlink() for symlinks and hard_link() for regular files, correctly recreating the entire symlink chain in the target directory


MarcusDunn:
not sure if merge broke, but CI need to be passing before I can merge. Thanks for the PR!

PR #1000, opened 2026-04-14
Update to latest version

marek-hradil:
Updated llama.cpp to `b8783`, so we can get the gemma4 audio support in 🎉 .
Related to that, there was a breaking change with `mtmd_decode_use_non_causal`, which now accepts a chunk parameter.
I have so far just instantiated to `null`, which should be okay, as it exactly mirrors the previous functionality.


MarcusDunn:
looks like this needs a fmt.

MarcusDunn:
you'll also have to rebase now I've merged https://github.com/utilityai/llama-cpp-rs/pull/995

marek-hradil:
@MarcusDunn It should be fixed now.

PR #1013, opened 2026-04-28
Add coverage and shell nix

mcharytoniuk:


MarcusDunn:
this is an unserious submission.

PR #1014, opened 2026-04-28
Allow dynamic backends

marek-hradil:
Adds a `dynamic-backends` Cargo feature that enables runtime hardware backend selection (Vulkan, CPU-AVX512, CPU-AVX2, etc.) so llama.cpp automatically picks the best available accelerator without any user configuration.

When enabled, the CMake build sets `GGML_BACKEND_DL=ON` and `GGML_CPU_ALL_VARIANTS=ON`, compiling each backend as a separate loadable module dropped into `OUT_DIR/backends/`; the path is propagated up the crate chain via `DEP_LLAMA_BACKENDS_DIR` → `cargo:rustc-env=GGML_BACKENDS_DIR` → `BACKENDS_DIR: Option<&str>` so downstream crates can locate it without their own build-script magic.


bogdanr:
This sounds like a very nice feature!

MarcusDunn (approved):

PR #1015, opened 2026-04-30
Add `common` feature flag to allow slimmer builds

bogdanr:
This PR introduces a new `common` feature flag to help downstream users who need to strictly constrain their binary sizes. 

By default, common is enabled, meaning existing users won't see any changes - the default Cargo build is byte-for-byte equivalent to how it was before.

### Why this matters:
If a user only needs the core inference API, they can now opt out of common. Doing so drops about ~24 MB of static archives from the link line (specifically libcommon.a at ~14 MB, the wrapper archive at ~10 MB, and their associated objects). 

### What gets disabled when common is off?

*    The OpenAI-compatible chat-template helpers
*    The JSON-schema-to-grammar converter
*    Lazy-grammar samplers
*    The streaming OAI parser

### Implementation details:

*    LlamaSampler::accept now gracefully falls back to llama.cpp's core llama_sampler_accept when the feature is off, so the basic sampling loop still works perfectly without the status-returning wrapper.
*    Examples relying on the OAI APIs (tools, tools_reasoning, openai_stream) now have required-features = ["common"], so Cargo will automatically skip them if you're doing a slim build.

Users who want the slim build can now just opt-in like this:
```
    llama-cpp-2 = { version = "0.1", default-features = false, features = [...] }
```

### Shameless plug?! 😄 
I am using this to make my project [fono](https://github.com/bogdanr/fono) compact and self contained.

### DIFF
--- a/llama-cpp-2/Cargo.toml
+++ b/llama-cpp-2/Cargo.toml
@@ -25,7 +25,13 @@ serde_json = "1.0"
 tracing-subscriber = { workspace = true }
 
 [features]
-default = ["openmp", "android-shared-stdcxx"]
+default = ["openmp", "android-shared-stdcxx", "common"]
+# Enable the OpenAI-compatible / chat-template / JSON-schema-to-grammar APIs.
+# These are backed by `llama.cpp`'s `common/` static library; disabling drops
+# `libcommon.a` (~14 MB) plus the `wrapper_common.cpp` / `wrapper_oai.cpp`
+# bindings (~10 MB combined) from the link line. Enabled by default for
+# backwards compatibility.
+common = ["llama-cpp-sys-2/common"]
 cuda = ["llama-cpp-sys-2/cuda"]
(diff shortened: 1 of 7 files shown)


MarcusDunn (approved):
this is great!

MarcusDunn:
I'll need fmt, lint, tests, passing on linux before I'll merge.

bogdanr:
Pushed a fmt fix in 37fd701, so Tests should be green next run.

The linux docker build is a separate problem that's already broken on main: same NCCL undefined symbols (`ncclCommInitAll` etc.) when building `simple --features cuda`. See the latest main run https://github.com/utilityai/llama-cpp-rs/actions/runs/26484680275 for the same trace. A ggml bump pulled in NCCL refs that the `cuda:12.3.1-devel` base image doesn't ship.

There's a fix waiting in run 26462504888 ("fix(cuda): disable NCCL autodetection by default + sync llama.cpp MTP wrappers/examples"), still parked on action_required. I'll rebase once that lands.

PR #1020, opened 2026-05-04
fix(cuda): disable NCCL autodetection by default + sync llama.cpp MTP wrappers/examples

MegalithOfficial:
## What this fixes

Some Linux CUDA builds were failing at link time on systems that had NCCL installed.

The errors looked like this:

```text
rust-lld: error: undefined symbol: ncclCommInitAll
rust-lld: error: undefined symbol: ncclGetErrorString
rust-lld: error: undefined symbol: ncclGroupStart
rust-lld: error: undefined symbol: ncclAllReduce
rust-lld: error: undefined symbol: ncclGroupEnd
collect2: error: ld returned 1 exit status
error: could not compile due to 1 previous error
```

The root cause was that `llama.cpp` could detect NCCL automatically during the CMake build, but `llama-cpp-sys-2` was not linking `libnccl` on the Rust side. That left us with compiled CUDA objects that referenced NCCL symbols but no final NCCL linkage.

## What changed

This update disables NCCL autodetection by default in CUDA builds by setting the relevant CMake flags in `llama-cpp-sys-2/build.rs`.

If someone does want NCCL enabled, they can still opt back in with:

```bash
LLAMA_CUDA_DISABLE_NCCL=0
```

## Why this approach

NCCL is mainly useful for multi-GPU collective operations. For normal single-GPU setups, auto-enabling it is unnecessary and makes builds more fragile on machines where NCCL happens to be installed.

This keeps the default CUDA build path reliable while still leaving a deliberate opt-in path for NCCL.

## Notes

If a previous build already cached NCCL detection, a clean rebuild may be needed.

Co-authored-by: Lothar Hoffmann

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1664,6 +1664,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "mtp-compare"
+version = "0.1.147"
+dependencies = [
+ "anyhow",
+ "clap",
+ "encoding_rs",
+ "hf-hub",
+ "llama-cpp-2",
+]
+
+[[package]]
 name = "mtp-example"
 version = "0.1.147"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1664,6 +1664,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "mtp-example"
+version = "0.1.147"
+dependencies = [
+ "anyhow",
+ "clap",
+ "encoding_rs",
+ "hf-hub",
+ "llama-cpp-2",
+]
+
+[[package]]
 name = "native-tls"
 version = "0.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
   "llama-cpp-2",
   "examples/embeddings",
   "examples/mtp",
+  "examples/mtp-compare",
   "examples/simple",
   "examples/reranker",
   "examples/mtmd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
   "llama-cpp-sys-2",
   "llama-cpp-2",
   "examples/embeddings",
+  "examples/mtp",
   "examples/simple",
   "examples/reranker",
   "examples/mtmd",
--- a/examples/mtp-compare/Cargo.toml
+++ b/examples/mtp-compare/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "mtp-compare"
+version = "0.1.147"
+edition = "2021"
+publish = false
+
+[dependencies]
+anyhow = { workspace = true }
+clap = { workspace = true, features = ["derive"] }
+encoding_rs = { workspace = true }
+hf-hub = { workspace = true }
+llama-cpp-2 = { path = "../../llama-cpp-2" }
+
+[features]
+cuda = ["llama-cpp-2/cuda"]
+metal = ["llama-cpp-2/metal"]
+vulkan = ["llama-cpp-2/vulkan"]
+
+[lints]
+workspace = true
--- a/examples/mtp-compare/src/main.rs
+++ b/examples/mtp-compare/src/main.rs
@@ -52,9 +52,14 @@ struct Args {
 
 #[derive(clap::Subcommand, Debug, Clone)]
 enum ModelSource {
-    Local { path: PathBuf },
+    Local {
+        path: PathBuf,
+    },
     #[clap(name = "hf-model")]
-    HuggingFace { repo: String, model: String },
+    HuggingFace {
+        repo: String,
+        model: String,
+    },
 }
 
 impl ModelSource {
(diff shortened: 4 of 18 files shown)


MegalithOfficial:
I added MTP support to the Rust wrapper and then wired up runnable examples on top of it. The wrapper now exposes the upstream MTP context type, recurrent-state config, and pre-norm embedding staging APIs needed for Qwen3.5-style NextN/MTP models. I also fixed mixed token+embedding batch handling on the Rust side, because upstream MTP needs both token ids and embedding rows in the same batch.

On top of that, I added two examples:
- `mtp-example`: generates with a bundled-MTP GGUF such as `unsloth/Qwen3.5-4B-MTP-GGUF`
- `mtp-compare`: runs plain decoding and MTP decoding back to back and prints output plus timing/acceptance stats

I also updated the MTP generation path to reuse live KV/recurrent state instead of clearing and reprefilling every loop. For Qwen3.5 MTP this required setting `n_rs_seq = draft_n` so rollback works correctly on the recurrent architecture.

Tested locally against `Qwen3.5-4B-Q4_K_M.gguf`. The path is working end to end; wall-clock speedup still depends on hardware, output length, and acceptance rate, so short CPU-only runs can still show MTP as slower even though the implementation is correct.

MegalithOfficial:
From what I understand, Action runs seem to suggest that Llama.cpp behaves differently across platforms: on Linux and macOS, the generated ctx_type binding comes through as one signedness, whereas on Windows it comes through as the other. I updated the wrapper to handle both, ensuring that the MTP context type code now compiles consistently on all targets.

MarcusDunn:
> From what I understand, Action runs seem to suggest that Llama.cpp behaves differently across platforms: on Linux and macOS, the generated ctx_type binding comes through as one signedness, whereas on Windows it comes through as the other. I updated the wrapper to handle both, ensuring that the MTP context type code now compiles consistently on all targets.

This is correct. bindgen generates the platform type for enums, your fix is correct.

MegalithOfficial:
I believe action error is now about i didnt do formatting.

MegalithOfficial:
Pushed a follow-up commit (`4c1da61`) that updates our `llama.cpp` submodule to the latest upstream commit (`39cf5d619` -> `ebc10770a`, build `b9611`, roughly 415 commits) and brings the wrappers back in line with the current API.

## What broke upstream

Three things in this range actually affected us.

Upstream renamed the `pre_norm` embeddings staging API to `nextn`. `llama_set_embeddings_pre_norm(ctx, value)` is now `llama_set_embeddings_nextn(ctx, value, masked)`, with matching getter renames. The new `masked` flag controls whether rows are stored only for output tokens (the old behavior) or densely for every token in the batch. Qwen3.5 MTP also extracts the post-norm hidden state now, so MTP output is worth a quick sanity run on a real model.

The mtmd helpers `mtmd_helper_bitmap_init_from_file/buf` changed signature. They now take a `placeholder` flag and return a small wrapper struct containing the bitmap plus an optional video context. This came with the new video input support, which is compile-gated behind `MTMD_VIDEO` — we leave it off, so there is no new ffmpeg dependency.

The CMake build gained a new `LLAMA_BUILD_APP` target (the unified `llama` binary) that defaults to ON for standalone builds and needs a generated `build-info.h`, which broke our build. `LLAMA_CURL` was also removed entirely; `common` now links a vendored `cpp-httplib` instead, optionally backed by system OpenSSL.

## What changed

- Renamed the `llama_rs_*_pre_norm` shims in `wrapper_common.{h,cpp}` to `llama_rs_*_nextn` and plumbed through the new `masked` parameter.
- Renamed the Rust side to match: `set_embeddings_nextn(enabled, masked)` and `embeddings_nextn[_ith]`. Passing `masked: true` reproduces the old indexing behavior exactly, and that is what the `mtp` and `mtp-compare` examples now do.
- Adapted `MtmdBitmap::from_file/from_buffer` to the new wrapper-struct return (passing `placeholder: false` to keep the old behavior). The bitmap now also stores the optional video context and frees it after the bitmap on `Drop`, so the wrapper stays correct if a future build enables `MTMD_VIDEO`.
- In `build.rs`: set `LLAMA_BUILD_APP=OFF` and `LLAMA_BUILD_UI=OFF`, replaced the removed `LLAMA_CURL=OFF` with `LLAMA_OPENSSL=OFF` (same policy: no system TLS/network dependencies), and added the vendored `cpp-httplib` static lib to the link line since `common` requires it now.
- Exposed the new `n_outputs_max` context param via `with_n_outputs_max` / `n_outputs_max`.

Everything else in the range was additive for us (`ctx_other`, `gguf_init_from_buffer`, lazy/placeholder bitmaps, chat parser fields), so `wrapper_oai.cpp` and the rest of the bindings needed no changes.

## To verify, I ran:

- `cargo build --workspace --features mtmd`
- `cargo test --workspace --features mtmd`
- `cargo check -p llama-cpp-2 --features mtmd,llguidance,sampler`
- `cargo fmt` + `cargo clippy --workspace --features mtmd`

Everything passed (91 tests, no new clippy warnings beyond the pre-existing pedantic set). I have not runtime-tested MTP generation against a real Qwen MTP model yet. Upstream's post-norm change makes that worth doing.

## Notes

If you have an existing local build, the cached CMake configuration predates the new options and the build will fail with `app/llama.cpp: fatal error: build-info.h`. A `cargo clean -p llama-cpp-sys-2` fixes it. Fresh builds (CI, crates.io consumers) are unaffected.

### AI Notice
I worked on this with help from Claude Fable 5 for the upstream diff analysis and the wrapper updates. I dont usually use AI for PRs but Fable 5 did the job in a way i would do and code seems to be fine.

MarcusDunn:
merge conflicts. otherwise looks good.

PR #1027, opened 2026-05-19
Add MTP speculative decoding support

DINOZYAVIER:
## Summary

Adds Rust bindings for llama.cpp MTP speculative decoding.

## Changes

- Updates llama.cpp to `ac7680`
- Adapts wrappers for moved common APIs
- Exposes MTP context params (`n_rs_seq`, context type)
- Adds `MtpSpeculative` safe wrapper

## Verification

- `cargo fmt --check`
- `git diff --check origin/main..HEAD`
- `cargo test -p llama-cpp-2`

### DIFF
--- a/llama-cpp-2/src/context/params.rs
+++ b/llama-cpp-2/src/context/params.rs
@@ -122,6 +122,37 @@ impl From<LlamaAttentionType> for i32 {
     }
 }
 
+/// A rusty wrapper around `llama_context_type`.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum LlamaContextType {
+    /// Default decoder context.
+    Default,
+    /// Multi-token-prediction draft context.
+    Mtp,
+    /// Unknown context type from a newer llama.cpp.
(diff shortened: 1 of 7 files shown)


MarcusDunn:
the crate version is already correct - otherwise the changes are welcome. I'll merge once version fixed and CI is clean.

DINOZYAVIER (commented):
Workflow test

DINOZYAVIER:
It looks like the workflow may have been cancelled due to a timeout. Before pushing, I built both targets that did not finish in CI: linux/amd64 completed fairly quickly, while arm64 took more than an hour.

DINOZYAVIER:
I resolved merge conflict and pushed. For context, previous CI attempts passed linux/amd64, macOS, Windows, and the Rust test job. Only `linux/arm64` was cancelled after about 6 hours during the Build step, which looks like a runner timeout.

madsmtm:
CC https://github.com/ggml-org/llama.cpp/issues/27469, would allow us to remove this glue code in the future.

Issue #1032, opened 2026-06-02
Cannot enable Intel MKL when building llama.cpp


fisher046:
# Summary

llama-cpp-sys-2 provides no way to build llama.cpp with the Intel MKL BLAS backend. There is no blas/mkl Cargo feature, and the build script's environment passthrough cannot reach the relevant CMake flags. Enabling MKL currently requires patching build.rs.

# Background

llama.cpp already supports Intel MKL. `ggml/src/ggml-blas/CMakeLists.txt` handles it explicitly.

# Details

No BLAS/MKL feature exists. The features in llama-cpp-sys-2/Cargo.toml are common, cuda, cuda-no-vmm, metal, vulkan, openmp, rocm, dynamic-link, *-stdcxx, system-ggml*, mtmd, and dynamic-backends. None enable a BLAS backend.

BLAS is effectively off everywhere. The only BLAS line in build.rs forces it off on Apple:

```rust
// llama-cpp-sys-2/build.rs:646
if matches!(target_os, TargetOs::Apple(_)) {
    config.define("GGML_BLAS", "OFF");
}
```

On non-Apple targets, GGML_BLAS is never set, so it falls back to llama.cpp's default, which is OFF for everything except Apple.

The env passthrough can't enable it. build.rs only forwards env vars prefixed with CMAKE_:

```rust
// llama-cpp-sys-2/build.rs:555
for (key, value) in env::vars() {
    if key.starts_with("CMAKE_") { config.define(&key, &value); }
}
```

The flags MKL needs — GGML_BLAS and GGML_BLAS_VENDOR — don't match that prefix, so there's no supported way to turn MKL on without editing the build script.

fisher046:
Hello @MarcusDunn, could you help look at the issue and my related PR? I see you are maintaining the repo. If you are not the right person, could you help find someone who can work on it? Thanks!

fisher046:
Resolved by https://github.com/utilityai/llama-cpp-rs/pull/1033.

PR #1033, opened 2026-06-03
Add a new Cargo feature mkl to compile llama.cpp with Intel MKL

fisher046:
The PR is related to the issue https://github.com/utilityai/llama-cpp-rs/issues/1032.

In some scenarios, especially for CPU inference, we want to compile llama.cpp with Intel MKL acceleration. llama.cpp already supports it. See: https://github.com/ggml-org/llama.cpp/blob/master/ggml/src/ggml-blas/CMakeLists.txt.

By adding the feature `mkl`, the compilation config `GGML_BLAS=ON` and `GGML_BLAS_VENDOR=Intel10_64lp`  can be added when the crate llama-cpp-sys compiles the llama.cpp.

### DIFF
--- a/examples/embeddings/Cargo.toml
+++ b/examples/embeddings/Cargo.toml
@@ -15,6 +15,7 @@ encoding_rs = { workspace = true }
 cuda = ["llama-cpp-2/cuda"]
 metal = ["llama-cpp-2/metal"]
 vulkan = ["llama-cpp-2/vulkan"]
+mkl = ["llama-cpp-2/mkl"]
 
 [lints]
 workspace = true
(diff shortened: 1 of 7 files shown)


MarcusDunn:
looks like this is a fmt away from being good to go. thanks for the PR @fisher046

fisher046:
@MarcusDunn I formatted the code. It seems that the failed WF job looks like not related to the PR.

fisher046:
@MarcusDunn Not sure if it looks good to you. Can you help review again?

MarcusDunn:
Apologies for dropping this. Thanks for the ping.

PR #1040, opened 2026-06-14
feat(build): allow configuring cmake parallelism via ENV

### LINKED ISSUE #993: no way to reduce build parallelism.
My laptop becomes pretty much unusable during build if max parallelism is used - I want to be able to run the build in the background, so I set `CMAKE_BUILD_PARALLEL_LEVEL` to `1` (acknowledging that the build will take longer).
However, it seems `build.rs` unconditionally overrides `CMAKE_BUILD_PARALLEL_LEVEL`, so this change has no effect. It seems there is actually no way to control the build parallelism.
I think if `CMAKE_BUILD_PARALLEL_LEVEL` is already set in the environment, it should not be overridden.

t3hmrman:
Resolves #993

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -1,8 +1,11 @@
-use cmake::Config;
-use glob::glob;
 use std::env;
+use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::str::FromStr as _;
+
+use cmake::Config;
+use glob::glob;
 use walkdir::DirEntry;
 
 enum WindowsVariant {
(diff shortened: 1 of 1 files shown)


MarcusDunn (approved):

t3hmrman:
Thanks for the review -- note that the error from the build looks unrelated:

```
 > [base-cuda 5/5] RUN cargo build --bin simple --features cuda:
4054.8           
4054.8           rust-lld: error: undefined symbol: ncclAllReduce
4054.8           >>> referenced by tmpxft_0000142d_00000000-6_ggml-cuda.compute_90.cudafe1.cpp
4054.8           >>>               ggml-cuda.cu.o:(ggml_backend_cuda_comm_allreduce_nccl(ggml_backend_cuda_comm_context*, ggml_tensor**)) in archive /target/debug/deps/libllama_cpp_sys_2-b8d637ca9c19e03c.rlib
4054.8           >>> referenced by tmpxft_0000142d_00000000-6_ggml-cuda.compute_90.cudafe1.cpp
4054.8           >>>               ggml-cuda.cu.o:(ggml_backend_cuda_comm_allreduce_nccl(ggml_backend_cuda_comm_context*, ggml_tensor**)) in archive /target/debug/deps/libllama_cpp_sys_2-b8d637ca9c19e03c.rlib
4054.8           collect2: error: ld returned 1 exit status
4054.8           
4054.8 
4054.8 error: could not compile `simple` (bin "simple") due to 1 previous error
```

PR #1048, opened 2026-06-17
feat: reintroduce api for tool use templating

vigsterkr:
unfortunately the latest releases dropped all support for tool use. this tries to add back some apis that supports tool use templating.

### DIFF
--- a/llama-cpp-2/src/common_chat.rs
+++ b/llama-cpp-2/src/common_chat.rs
@@ -1,0 +1,9 @@
+//! llama.cpp common-chat template and parser utilities.
+
+pub use crate::model::{
+    ChatTemplateResult as CommonChatTemplateResult, GrammarTrigger, GrammarTriggerType,
+};
+pub use crate::openai::{
+    ChatParseStateOaicompat as CommonChatParseState,
+    OpenAIChatTemplateParams as CommonChatTemplateParams,
+};
--- a/llama-cpp-2/src/lib.rs
+++ b/llama-cpp-2/src/lib.rs
@@ -23,6 +23,8 @@ use std::os::raw::c_int;
 use std::path::PathBuf;
 use std::string::FromUtf8Error;
 
+#[cfg(feature = "common")]
+pub mod common_chat;
 pub mod context;
 pub mod gguf;
 pub mod llama_backend;
@@ -33,6 +35,8 @@ mod log;
 pub mod model;
 #[cfg(feature = "mtmd")]
 pub mod mtmd;
+#[cfg(feature = "common")]
+pub mod openai;
 pub mod sampling;
 pub mod timing;
 pub mod token;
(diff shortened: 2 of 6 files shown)


vigsterkr:
@MarcusDunn @AsbjornOlling i wonder if you guys would consider adding this back, or do i need to keep maintaining a fork? since inference should include tool use as well.
reasoning_content is something that wasn't working for me correctly before these changes.

AsbjornOlling:
To start: I'm okay with going either way.

My thinking is that the goal of llama-cpp-rs is to be "dead simple, very close to raw bindings". I think adding hundreds (or thousands) of lines of C++ is contrary to the goal being "dead simple", and that adding json handlilng and API stuff moves it away from being low-level bindings to llama.cpp.

I think that the valuable stuff to extract from llama.cpp is the different model architectures implemented, the hardware acceleration backends from ggml, stuff like that.

Handling chat templates, json parsing, tool call parsing and stuff like that is in my opinion best done in rust, outside of llama.cpp. It's all very high-level features, and I think this lib is best served by being relatively low-level, and operating on the level of logits and tokens, rather than on the level of messages.

That's what we're doing in the [NobodyWho](https://github.com/nobodywho-ooo/nobodywho/) project. E.g. we use [minijinja](https://github.com/mitsuhiko/minijinja) to handle chat templating, and implement the tool calling parsers ourselves.

Originally, this library was explicit about not handling any of the stuff in `common`, focusing only on the dependency-free stuff in libllama. This scope has expanded over time (IIRC mtmd triggered this), and I think it brings on a lot of complexity. More build-time deps, more C++ code, etc.

It looks like the openai api stuff will be a bit more difficult to maintain in the future, since some of these things started moving out of common and into server in the upstream.

@vigsterkr is it feasible for you to reimplement the openai json stuff you need in rust? If not, I'm very okay with reintroducing some of it in this lib, although I prefer for this lib to get closer to the original low-level scope.

vigsterkr:
@AsbjornOlling thnx for getting back! my thinking was that many of the things for templating and tool call handling etc is already implement in llama.cpp (for obvious reasons). now if i start implementing things on rust side it would be basically the re-implementation of what is done in llama.cpp hence i thought to expose those binding to rust and try to do it on c++ level using the native code that is being maintained by the llama.cpp dev team. i think they know better how to handle this then a binding crate/library.

AsbjornOlling:
@vigsterkr 
> i think they know better how to handle this then a binding crate/library.

I wouldn't be too sure about that.

E.g. the chat template stuff currently in llama.cpp is a lot slower than minijinja: https://gitlab.com/AsbjornOlling/chat-template-benchmark. 

Same thing goes for grammar sampling. LLGuidance, a pure-rust crate, is significantly faster (and more featureful) than the grammar samplers implemented in llama.cpp: https://github.com/guidance-ai/jsonschemabench/tree/main/maskbench
I have also personally come across more than a few bugs with the llama.cpp grammar sampling. String-based lazy grammars are extremely inefficient (they scale with the length of the full context). The GBNF parser was crashing for a while whenever you didn't have a rule literally called "root" (I fixed this in upstream PR 19761), and I think there's still a serious bug in upstream that causes a segfault with a significant subset of valid grammars.

When llama.cpp does context shifting, it discards stuff based on an token-count offset:
https://github.com/ggml-org/llama.cpp/blob/e2e7a9b2d04e0404131e551243d4bd399ccfe606/tools/server/server-context.cpp#L2570
This means that it will with a high probability leave behind a garbled context, with messages split mid-sentence, or with the entire chat format messed up (messages missing eos tokens, or two identical roles in a row, etc etc). Compare this to NobodyWho, where we only context-shift along message boundaries, making sure that the context still makes sense after shifting.

TL;DR: I don't think you should put too much faith in the high-level features of llama.cpp being optimal.

MarcusDunn:
I think tool calling can be handled in rust as @AsbjornOlling described much better than going though bindings to llama.cpp. 

I no longer use this library so I will continue to be pretty deferential, but having implemented tool calling on top of this lib back for my own use-case when it had a considerably smaller API surface, my instinct is not to merge this.

PR #1049, opened 2026-06-19
Disable Metal backend on watchOS (no Metal framework)

s194042:
Problem: watchOS has no Metal framework, so ggml-metal's find_library(Metal) aborts CMake configure and the build script panics.  
Fix: detect watchOS as its own AppleVariant, define GGML_METAL=OFF, add -D_DARWIN_C_SOURCE, and skip linking the Metal/MetalKit frameworks.


MarcusDunn:
great fix. thanks for the PR.

PR #1050, opened 2026-06-21
Add bindings for the 3 missing KV-cache functions

sdroege:


sdroege:
I'm planning to use this in https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/3119 for more efficient KV-cache reuse similar to the "infinite text generation" in https://github.com/ggml-org/llama.cpp/blob/bddfd2b1137cd6e51fbb939081caf50e9f496a66/tools/completion/completion.cpp#L604

MarcusDunn:
thanks for the PR. looks good to me.

PR #1056, opened 2026-06-24
MTP speculative decoding

replikeit:
Binds llama.cpp's already-implemented `draft-mtp` speculative decoding into the Rust crates, so downstream code can run an MTP speculative loop with a single self-contained GGUF (embedded NextN head — no separate draft model).

This is **binding work only**: the C++ MTP graph and draft logic already exist in the pinned vendored tree (`9e3b928`). No submodule bump, no C++ inference/graph changes.

## What's added

**`llama-cpp-sys-2`** — a flat `extern "C"` shim (`llama_rs_speculative_*`) over libcommon's `common_speculative` API, compiled under the existing default `common` feature (no new feature flag). C++ vectors and the reference-returning `common_speculative_get_draft_params` are flattened to C arrays + status codes; all exceptions are caught at the boundary and mapped to `llama_rs_status`.

**`llama-cpp-2`**
- `LlamaContextType { Default, Mtp }` + `with_context_type()` / `context_type()` builder/getter.
- `with_n_rs_seq()` / `n_rs_seq()` — recurrent-state rollback snapshots, required for partial `seq_rm` on recurrent/hybrid archs.
- `LlamaModel::new_mtp_context()` — creates the draft context on the **same model** (`ctx_type = Mtp`, `ctx_other = target`), mirroring `tools/server/server-context.cpp`.
- `MtpSpeculator` — safe-ish wrapper owning the shim handle: `begin` / `process` / `draft` / `accept` / `need_embd_nextn`. Acceptance stays in Rust (`common_sampler_sample_and_accept_n` is intentionally **not** shimmed), so callers keep their own sampling/grammar/streaming.

**`examples/speculative-mtp`** — a `clap` binary that drives the loop, prints generated text + `accept%`, and has a `--baseline` greedy mode for comparison. Doubles as a smoke test. Plus an `#[ignore]` integration test (`LLAMA_MTP_MODEL`-gated).

**`build.rs`** — sets `GGML_CUDA_NCCL=OFF` under the `cuda` feature. NCCL only drives multi-GPU collectives this single-process crate never uses; upstream's `ON` default made `ggml-cuda` reference `nccl*` symbols the static link line never resolved, breaking `cargo build --features cuda` on any host with NCCL installed.

## Safety notes

- `MtpSpeculator::new` is `unsafe`: it stores raw pointers to the target/draft contexts inside llama.cpp; both must outlive the speculator. A shared borrow can't express this (the contexts are still mutated via `decode`).
- `LlamaModel::new_mtp_context` is likewise `unsafe`: the returned draft context holds a raw `ctx_other` pointer into `target`, whose lifetime isn't bound by the borrow checker — `target` must outlive the result.

## Validation

Verified end-to-end on `Qwen3.5-4B-UD-Q4_K_XL.gguf` (arch `qwen35`, hybrid Gated DeltaNet + MoE):

| backend            | baseline   | MTP        | speedup | accept% |
|--------------------|------------|------------|---------|---------|
| CPU                | 8.4 tok/s  | 11.7 tok/s | 1.39×   | 50.0%   |
| CUDA (RTX 4080)    | 112 tok/s         | 178.0 tok/s| 1.62x       | 44.6%   |

- `accept% > 0` confirms the embedded NextN head drafts useful tokens.
- Recurrent rollback (`seq_rm`) succeeds on the qwen35 arch with `n_rs_seq = n_draft + 4`.
- CUDA build links cleanly with the NCCL fix — no `RUSTFLAGS` workaround needed.

CI green locally:

```bash
cargo fmt --check
cargo clippy
cargo test --features sampler   # 18 unit + 92 doctests pass; integration test #[ignore]
```

## Non-goals

- No EAGLE3 / n-gram speculative wiring.
- No multi-sequence / batched-server orchestration (single-sequence loop; the shim takes `n_seq` so it can generalize later).
- No C++ inference/graph changes — upstream MTP used as-is.


replikeit:
Didn't noticed that feature was merged 19h ago, sorry ;))))))))))))))))))))))))))

PR #1058, opened 2026-07-02
Model free gbnf validator

mcharytoniuk:


mcharytoniuk:
sry wrong origin, sorry

PR #1066, opened 2026-07-08
feat: add llguidance_with_slices to supply custom slice regexes

jona605a:
### Summary

`LlamaSampler::llguidance` always builds its ParserFactory with `ParserFactory::new_simple`, which pre-computes bitmasks for four general JSON regexes. During inference, whenever only tokens in a slice could be valid, llguidance uses one of these pre-built bitmasks instead of walking the full vocabulary trie. However, this optimization only happens if the slice matches the token sub-set the grammar actually needs at that position.

For non-JSON grammars, the built-in JSON slices do essentially nothing and can waste a lot of set-up time. This PR exposes a variant that lets callers supply their own slice regexes.

### Changes

`llguidance_sampler.rs`
- Refactored the tail of `create_llg_sampler` into a private `finish_llg_sampler` helper shared by both functions
- Added `create_llg_sampler_with_slices(model, grammar_kind, grammar_data, slices)` which builds `ParserFactory::new(tok_env, InferenceCapabilities::default(), slices)` instead of `new_simple`

`sampling.rs`
- Added `LlamaSampler::llguidance_with_slices(model, grammar_kind, grammar_data, slices)` as the public API counterpart to the existing `LlamaSampler::llguidance`

### Usage
```
// JSON-heavy grammars — four generic JSON regexes
let slices = llguidance::earley::SlicedBiasComputer::general_slices();

// Delimiter-bounded formats (e.g. Gemma-4 tool calls)
let slices = vec![r"[^<>{},:]+".to_string()];

// Disable the optimization entirely
let slices: Vec<String> = vec![];

LlamaSampler::llguidance_with_slices(model, "lark", grammar, &slices)?
```

### DIFF
--- a/llama-cpp-2/src/llguidance_sampler.rs
+++ b/llama-cpp-2/src/llguidance_sampler.rs
@@ -4,10 +4,10 @@
 //! to enforce grammar constraints (JSON schema, regex, Lark, etc.) during token sampling.
 
 use std::ffi::c_void;
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 
 use llguidance::Matcher;
-use toktrie::{ApproximateTokEnv, TokRxInfo, TokTrie};
+use toktrie::{ApproximateTokEnv, TokEnv, TokRxInfo, TokTrie};
 
 use crate::model::LlamaModel;
 use crate::sampling::LlamaSampler;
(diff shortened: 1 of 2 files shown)


AsbjornOlling:
It might be neat to expose a more generic way of creating llama.cpp samplers from llguidance samplers, so llama-cpp-rs doesn't explode in scope to expose all possible llguidance configurations.

MarcusDunn:
I think we should steer users away from using the llama-cpp llguaidance altogether. guidance can be layered on top of existing APIs (and llguidance is rust anyways)

MarcusDunn:
I'm going to close this - I would like to decrease our llguidance api surface to the point of removing it instead of adding to it.

I appreciate the PR though.

jona605a:
Yes, I saw your feedback and am planning on closing this and opening a new PR tomorrow, cutting down on the llama-cpp-rs api while opening it up (Asbjorn's comment) to accepting a caller-created Matcher. I'd love to receive your feedback on it.

PR #1068, opened 2026-07-09
Adding safe wrappers for seq_state functions

s194042:
For hybrid models like Qwen3.5 we need to be able to save and restore checkpoints of the sequence state. The current flow is:

1. Get size of seq state.
2. Create a mutable buffer of that size. Pass the seq id, a raw pointer to your buffer and SeqFlags to state_seq_get_data_ext. This directly calls FFI C function of the same name, however passes usize:MAX as the buffer size. This is unsafe as there are no guarantees about the buffer and the llama.cpp is not told the actual buffer size. 
3. In case of a restore, you pass a seq id, a Vec<u8> containing your checkpoint and the same flags you used when creating the checkpoint to state_seq_set_data_ext, which passes these to the FFI C function. This is quite unsafe, because the parser in llama.cpp cannot trust the bytes passed to state_seq_set_data_ext.

We attempt to address these safety concerns by introducing a new getter and setter for checkpoints:
`state_seq_get` and `state_seq_set` and new type SeqState.

The new type, SeqState, stores the bytes of a checkpoint and the flags used to create them. This becomes the return of the safe getter and the data input of the safe getter. This way we ensure that the new setter only receives bytes created by llama.cpp and greatly increase the trustworthiness of the data. There are many checks worth making on the bytes generated by llama.cpp (You can not use a checkpoint from one model family with another), but these are handled by llama.cpp internally. Keeping the flags in SeqState also removes the option of trying to restore a checkpoint with the incorrect flags.
The new getter now handles step 1. and 2. at the same time and also passes the correct buffer size to the C function, which means the caller is not responsible for the buffer.

Additionally we also update the LLamaStateSeqFlags with ON_DEVICE so it is possible to keep the checkpoint on GPU and implement OR'ing, so multiple flags can be passed. 

We leave the preexisting unsafe functions as they allow for more flexibility.


MarcusDunn:
needs a `fmt`

s194042:
> needs a `fmt`

done :)

MarcusDunn:
will merge on CI pass. please ping if I drop this.

PR #1069, opened 2026-07-12
Enable Sampler to backend offloading

ju6ge:
Many of the standard sampling steps like `min-p` and `temp` have gpu optimized 
versions availible. Until now the rust bindings did not provide the right 
entrypoints to enable offloading for supported samplers (support may vary depending on the backend type as well!). Until now all samplers where run on the cpu side. Depending on the sampling chain topology this could lead to excessive pcie transfers. This is also highly model dependent, but especially the gemma model have a huge set of possible tokens -> many logics that need to cross the pcie bus. Even relativly small chains of samplers could deteriorate performance measurably!

This PR enables adding samplers at context creation time. More complex samplers 
like repetition penilties or grammars can not be run on any backend! 

Backend enabled samplers run on context decode. You can also only do partial steps on the backend, for example:

```rust,ignore
let sampler = LlamaSampler::chain([
    LlamaSampler::min_p(0.01, 64),
    LlamaSampler::temp(0.1),
], false);


let final_select = LlamaSampler::chain([
    // other cpu only samplers
    LlamaSampler::dist(),
], false);

let ctx = model.new_context_with_samplers(
    &backend,
    ctx_params,
    [(0, sampler)],
)?;

// Initialize prompt tokens and run

ctx.decode(prompt_tokens)?;

// sample next token
let token = final_select.sample(&mut ctx, -1);
```

I have successfully integrated this into my downstream project and am very pleased 
with the improved flexibility and performance. Have fun with it ;)

Cheers 
ju6ge

### DIFF
--- a/llama-cpp-2/src/context.rs
+++ b/llama-cpp-2/src/context.rs
@@ -7,6 +7,7 @@ use std::slice;
 
 use crate::llama_batch::LlamaBatch;
 use crate::model::{LlamaLoraAdapter, LlamaModel};
+use crate::sampling::LlamaSampler;
 use crate::timing::LlamaTimings;
 use crate::token::data::LlamaTokenData;
 use crate::token::data_array::LlamaTokenDataArray;
@@ -28,6 +29,8 @@ pub struct LlamaContext<'a> {
     pub model: &'a LlamaModel,
     initialized_logits: Vec<i32>,
     embeddings_enabled: bool,
+    /// Backend samplers kept alive for the context's lifetime.
+    _backend_samplers: Vec<(i32, LlamaSampler)>,
 }
 
 impl Debug for LlamaContext<'_> {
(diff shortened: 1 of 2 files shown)


llama-cpp-2/src/context.rs
```diff
@@ -28,6 +29,8 @@ pub struct LlamaContext<'a> {
     pub model: &'a LlamaModel,
     initialized_logits: Vec<i32>,
     embeddings_enabled: bool,
+    /// Backend samplers kept alive for the context's lifetime.
+    _backend_samplers: Vec<(i32, LlamaSampler)>,
```
MarcusDunn, line 33: as mutating `_backend_samplers` is illegal (from my understanding?), I suspect a Box<[(i32, LlamaSampler)]> may be a better type to enforce that internally.

(internal) documentation on what you can and cannot do with these safely would also be appreciated.

MarcusDunn (changes requested):
great pr. I like the additon. see comment for small nit.

llama-cpp-2/src/context.rs
ju6ge: Yes mutating the samplers is not something that should be done … Llama.cpp has as set_samplers method for ctx, but in my testing that only lead to crashes. Backend samplers are wired into the compute graph and updating them in any way after initialization is highly experimental … At least when trying to update a sampler for an existing sequence. Not sure what happens if multiple sequences use different sampling chains.

But would Box<[(i32, LlamaSampler)] really be better? I mean this field is not exposed on the downstream side of things, so any manipulation would need to happen via a LlamaContext method anyway.

ju6ge:
Thanks, sorry for not having gotten back to you, other projects have kept me busy ;)

Regarding the comment on documentation: What kind of information would be useful. I would like to avoid explaining a moving target in terms of backend <-> sampler support. Any docs about which sampler is enabled on which backend is likely to go out of date quite quickly. Optimizing the sampling pipeline using this is bound to be quite specific on what the goal is.

I could expand the docstring on `new_context_with_samplers` a bit more, is that the right spot?

PR #1072, opened 2026-07-14
feat: simpler llguidance API and more customizable samplers

jona605a:
Replaces #1066 , as the scope changed after review. 

The aim of this PR is to reduce the llguidance API surface of llama-cpp-rs, while giving callers the option to create more customizable samplers. 

What changed
- Added `From<Matcher>` for `LlamaSampler` as the primary way to support llguidance
- Removed `llguidance()` and `create_llg_sampler`. Old callers must now build their own `ParserFactory` / `Matcher` and convert.
- Exposed `llguidance_tok_env` as the only internal llama-cpp-rs function callers need before converting - needed because it has internal access to the model's vocabulary

The struct `LlgContext` only used the `Matcher` field, so it was removed, and the functions for C adaptation now points to `Matcher` instead of `LlgContext`. 

To help users whose calls to llguidance will now error, explanatory doc comments have been added and examples/llguidance.rs has been updated. 

For verification, the example still gives the same output as before this PR.

### DIFF
--- a/examples/llguidance.rs
+++ b/examples/llguidance.rs
@@ -13,6 +13,8 @@ use llama_cpp_2::model::params::LlamaModelParams;
 use llama_cpp_2::model::AddBos;
 use llama_cpp_2::model::LlamaModel;
 use llama_cpp_2::sampling::LlamaSampler;
+use llguidance::api::TopLevelGrammar;
+use llguidance::{Matcher, ParserFactory};
 use std::io::Write;
 
 fn main() {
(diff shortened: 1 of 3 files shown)


MarcusDunn:
I like this PR much more. some discussion on the saftey of the casts would be appreciated (ik they existed before, but the type changing is a good excuse to improve on things)

jona605a:
Thanks for the feedback. The safety of the casts are a bit out of scope, so I reverted that change.

jona605a:
Any other comments, @MarcusDunn ?

PR #1084, opened 2026-07-28
Update build.rs to handle Vulkan android builds

s194042:
There is no Android arm for the Vulkan section of build.rs. This adds the Android arm and handles the env variables and headers that the Android NDK cannot find itself. 

Additionally it consolidates the Android toolchain selection to lessen code duplication.


AsbjornOlling:
@MarcusDunn ping :sparkles:

PR #1087, opened 2026-08-02
fix: size embeddings reads by the LIVE pooling type, not n_embd

thermodynamic-retard:
## What

`LlamaContext::embeddings_seq_ith` and `embeddings_ith` build their return slice with `self.model.n_embd()`. That is the wrong length for every pooling mode:

| mode | correct width | why |
|---|---|---|
| `RANK` | `n_cls_out` (default 1) | llama.h:1029 sizes `llama_get_embeddings_seq` as `float[n_cls_out]` — the `n_embd`-wide read **over-reads** |
| `MEAN`/`CLS`/`LAST` | `n_embd_out` | llama-context.cpp's extraction switch resizes each `embd_seq` to `n_embd_out`, which diverges from `n_embd` whenever `{arch}.embedding_length_out` is present (deepstack models) |
| `NONE` (per-token) | `n_embd_out` | `get_embeddings_ith` returns `embd.data + j*n_embd_out`; the switch copies `n_tokens*n_embd_out` floats for NONE too |

Creating the over-long slice violates `slice::from_raw_parts`' validity contract (latent UB). The crate's own `examples/reranker` reads `[0]` of a RANK sequence and "works" only because the over-read lands in the same allocation.

## Fix

Size the read from the context's **LIVE pooling type** via a new `embeddings_out_len()` helper: `llama_model_n_cls_out` for RANK, `llama_model_n_embd_out` otherwise. Two new `LlamaModel` accessors (`n_embd_out()`, `n_cls_out()`) mirror the existing `n_embd()`.

## Notes

- Ground-truth verified against llama.cpp 0.1.150: `get_embeddings_ith` returns `embd.data + j*n_embd_out` and the extraction switch sizes `embd_seq` by pooling (`llama-context.cpp:1440-1480`); llama.h:1029 states the RANK width.
- This is a draft — CI is the first full compile of the change. omm (a downstream user of a vendored fork of this crate) carries the identical fix and its green build is the semantic proof.

Closes the S6-UB over-read behind omm's fork.

### DIFF
--- a/llama-cpp-2/src/context.rs
+++ b/llama-cpp-2/src/context.rs
@@ -140,7 +140,9 @@ impl<'model> LlamaContext<'model> {
     /// # Returns
     ///
     /// A slice containing the embeddings for the last decoded batch.
-    /// The size corresponds to the `n_embd` parameter of the context's model.
+    /// The size is the pooling-derived output width: `n_cls_out` for RANK,
+    /// `n_embd_out` otherwise — NOT `n_embd` (llama.h:1029 /
+    /// llama-context.cpp's extraction switch).
     ///
     /// # Errors
     ///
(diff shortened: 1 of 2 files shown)


MarcusDunn (approved):
lgtm. thanks for the PR!

PR #1093, opened 2026-08-08
sampling: catch C++ exceptions from sample/apply at the FFI boundary

### LINKED ISSUE #1082: Grammar sampler: C++ exception during sampling crosses the FFI boundary and aborts the process
#874 added `GrammarError` and input sanitizing so invalid grammars fail with an error instead of an abort at construction time. The same hazard still exists at sampling time: if the grammar sampler ends up with an empty candidate set during generation, llama.cpp throws a C++ exception inside `llama_sampler_apply`/`accept`, the exception unwinds across the FFI boundary, and the process dies with:

```
fatal runtime error: Rust cannot catch foreign exceptions, aborting
```

There is no way for downstream Rust code to catch or handle this.

Real-world case: nobodywho-ooo/nobodywho#421. Their JSON preset built a chain of `[top_k(20), top_p, temp, grammar]`. With models whose top-k candidates contain no grammar-valid token (thinking models like Qwen3-0.6B, which want to open with `<think>`), the grammar masked every candidate and the process aborted silently inside the Godot editor. They fixed the ordering (nobodywho-ooo/nobodywho#644), but the underlying hazard remains for any chain where the candidate set can go empty, and it reproduces through plain `LlamaSampler` usage: `chain([top_k(1), grammar("root ::= \"some-unlikely-literal\"")])` plus a sampling loop aborts with any model.

Possible directions, in order of preference:

1. Catch the exception on the C++ side of llama-cpp-sys-2 (a small shim wrapping `llama_sampler_apply`/`llama_sampler_accept` in try/catch) and surface it as an error or a poisoned-sampler state, similar to what #874 did for construction.
2. At minimum, document the hazard on `LlamaSampler::grammar`/`grammar_lazy` so chain ordering mistakes fail loudly in review instead of in production.

Happy to work on a PR for either direction if there is interest.

GuiBarradas:
Closes #1082

llama.cpp can throw a C++ exception during sampling, for example when a grammar masks every remaining candidate (a grammar sampler placed after a truncation sampler such as top-k). The exception unwinds across the FFI boundary, Rust cannot catch foreign exceptions, and the process aborts:

```
fatal runtime error: Rust cannot catch foreign exceptions, aborting
```

#874 added try/catch wrappers for grammar construction and `accept`. This extends the same pattern to the sampling path:

- C++ shim (`wrapper_common.cpp/.h`): `llama_rs_sampler_sample` and `llama_rs_sampler_apply`, wrapping `llama_sampler_sample` / `llama_sampler_apply` in try/catch and returning `llama_rs_status`, mirroring the existing `llama_rs_sampler_accept`.
- Rust: `LlamaSampler::try_sample` and `LlamaTokenDataArray::try_apply_sampler`, returning a new `SamplerSampleError` instead of aborting. Gated on the `common` feature, like `try_accept`. The existing `sample`/`apply` are unchanged.

Verified on Windows (CPU build): the crate compiles, and a chain of `[top_k(1), grammar("root ::= \"unlikely\""), dist]` that aborted the process via `sample` now returns `Err(SamplerSampleError::FfiError(-3))` via `try_sample`, with the process surviving. This is the same failure real callers hit (nobodywho-ooo/nobodywho#421, whose generation loop uses `sample`).

No tests added, since the failure needs a loaded model and the crate's unit tests do not use one. Happy to add one under a feature/env gate if you prefer.

### DIFF
--- a/llama-cpp-2/src/sampling.rs
+++ b/llama-cpp-2/src/sampling.rs
@@ -11,6 +11,8 @@ use crate::status_is_ok;
 use crate::token::data_array::LlamaTokenDataArray;
 use crate::token::logit_bias::LlamaLogitBias;
 use crate::token::LlamaToken;
+#[cfg(feature = "common")]
+use crate::SamplerSampleError;
 use crate::{GrammarError, SamplerAcceptError};
 
 /// A safe wrapper around `llama_sampler`.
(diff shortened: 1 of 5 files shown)


llama-cpp-2/src/sampling.rs
```diff
@@ -35,6 +37,40 @@ impl LlamaSampler {
         LlamaToken(token)
     }
```
madsmtm, line 38: Consideration: Maybe we should use the same pattern in here as in `LlamaSampler::accept`? That is, to call `self.try_sample().unwrap()` when `feature = "common"`, and fall back to `llama_sampler_sample` when not?

madsmtm (approved):
In general, I tend to think that we should use `extern "C-unwind"` instead of writing our own wrappers. For now though, I think this is probably fine.

madsmtm:
I have also opened https://github.com/ggml-org/llama.cpp/discussions/28085 to try to figure out how `llama.cpp` wants us to handle this.

GuiBarradas:
Thanks for the review. You are right that `extern "C-unwind"` is the cleaner long-term direction, and I am happy to switch to it if that is what you prefer. I used the explicit C++ try/catch wrappers here mainly to match the existing `llama_rs_sampler_accept` pattern in wrapper_common.cpp, so the sampling path is handled the same way as accept. Happy to follow whatever comes out of the llama.cpp discussion you opened.

PR #1104, opened 2026-08-20
Redirect logs from `mtmd` to `tracing`

madsmtm:
`mtmd.h` and `mtmd-helper.h` have their own logging statics, so we need to configure those in addition to the existing configuration for `llama.h` and `ggml.h`.


AsbjornOlling:
@MarcusDunn can we merge this? :innocent:

madsmtm:
I recently joined NobodyWho as well FYI (which is why Asbjørn is asking), I guess you can see from the PR fork name, but I should probably have made it clearer that I'm writing these PRs for them.

MarcusDunn (approved):

PR #1109, opened 2026-08-25
Fix building `llguidance` with newer `llama.cpp`

madsmtm:
I think this was introduced in https://github.com/utilityai/llama-cpp-rs/pull/1097, the glue code for `llguidance` wasn't updated when `llama.cpp` was.

I've also updated the CI workflow to prevent issues like this in the future.

### DIFF
--- a/.github/workflows/llama-cpp-rs-check.yml
+++ b/.github/workflows/llama-cpp-rs-check.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Fmt
         run: cargo fmt --check
       - name: Test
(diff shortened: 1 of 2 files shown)


MarcusDunn:
Needs a fmt. Doesn't look like the problem was introduced here - but I'd like checks to run.

MarcusDunn:
thanks! will merge provided CI is green

madsmtm:
I've opened https://github.com/utilityai/llama-cpp-rs/pull/1110 to fix the CI

PR #1112, opened 2026-08-27
Simplify the lookup for `GGML_BACKENDS_DIR`

madsmtm:
Removes a build script, which should make compilation slightly faster.

### DIFF
--- a/llama-cpp-sys-2/build.rs
+++ b/llama-cpp-sys-2/build.rs
@@ -1064,6 +1064,13 @@ fn main() {
         config.define("GGML_CPU_ALL_VARIANTS", "ON");
         config.define("GGML_BACKEND_DIR", backends_dir.to_str().unwrap());
         // BUILD_SHARED_LIBS=ON is already set above via the dynamic-link feature.
+        println!(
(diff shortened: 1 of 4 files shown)


llama-cpp-sys-2/build.rs
```diff
@@ -1064,6 +1064,13 @@ fn main() {
         config.define("GGML_CPU_ALL_VARIANTS", "ON");
         config.define("GGML_BACKEND_DIR", backends_dir.to_str().unwrap());
         // BUILD_SHARED_LIBS=ON is already set above via the dynamic-link feature.
+        println!(
+            "cargo:rustc-env=GGML_BACKENDS_DIR={}",
+            backends_dir.display()
+        );
+
+        // Expose backends dir to direct dependencies as `DEP_LLAMA_BACKENDS_DIR`.
+        println!("cargo:backends_dir={}", out_dir.join("backends").display());
```
madsmtm, line 1073: I see that https://github.com/nobodywho-ooo/nobodywho/pull/691 is actually using this, and it might not be the only one. So I've opted to not remove this just yet.

AsbjornOlling (approved):

PR #1122, opened 2026-09-03
Soundness: Fix invalid mutation behind `&`

madsmtm:
The `llama.cpp` API is actually fairly Rusty, and communicates pretty well which function parameters are mutated (either the parameter is `const`, and it's not mutated, or it's not `const`, and it's potentially mutated).

We can take advantage of this information by introducing a helper `Ptr` with methods `as_ptr(&self) -> *const T` and `as_mut_ptr(&mut self) -> *mut T` to access the underlying pointer, which then ensures that when the API requires a `*mut` pointer, it came from a `&mut`.

This also removes an incorrect `Clone` impl on `MtmdBitmap` that would lead to a double-free and/or use-after-free.

### DIFF
--- a/examples/llguidance.rs
+++ b/examples/llguidance.rs
@@ -75,7 +75,7 @@ fn main() {
     std::io::stdout().flush().unwrap();
 
     while n_cur <= 128 {
-        let token = sampler.sample(&ctx, batch.n_tokens() - 1);
+        let token = sampler.sample(&mut ctx, batch.n_tokens() - 1);
 
         if token == model.token_eos() {
             break;
--- a/examples/mtmd/src/mtmd.rs
+++ b/examples/mtmd/src/mtmd.rs
@@ -181,7 +181,7 @@ impl<'a> MtmdCliContext<'a> {
         // Clear bitmaps after tokenization
         self.bitmaps.clear();
 
-        self.n_past = chunks.eval_chunks(&self.mtmd_ctx, context, 0, 0, batch_size, true)?;
+        self.n_past = chunks.eval_chunks(&mut self.mtmd_ctx, context, 0, 0, batch_size, true)?;
         Ok(())
     }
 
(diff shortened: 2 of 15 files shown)


madsmtm:
Full list of API changes:
- A bunch of `LlamaContext::*` methods now takes `&mut self`. This is correct, these methods call `ctx->synchronize()` internally, which mutates various context parameters (and is definitely not thread-safe).
- `LlamaSampler::sample` now takes `&mut LlamaContext`. This also synchronizes the context.
- `LlamaModel::lora_adapter_init` now takes `&mut self`. This is correct since loras are added to a set on the model, and removed from that when the lora is destroyed.
- `MtmdBitmap::set_id` now takes `&mut self`. This is clearly a mutating method.
- `MtmdContext::encode_chunk` now takes `&mut self`. This is correct since the context's clip is written to during this operation.
- `MtmdInputChunks::eval_chunks` now takes `&mut MtmdContext` and `&mut LlamaContext`. Similar to above, this mutates the contexts when decoding, and is the reason why the function isn't thread-safe.
- Removed `impl Clone for MtmdBitmap`.
- `LlamaSampler::apply` and `LlamaTokenDataArray::apply_sampler` now takes a mutable reference to the sampler. This is required for some samplers such as mirostat_v2.

I can split it into separate PRs if you'd prefer that?

madsmtm:
There is another option here: that we treat `LlamaContext` as-if it uses interior mutability (e.g. as-if all of its fields were wrapped in a `Cell` or `RefCell`). This makes it impossible to make `Send + Sync`, which is unfortunate, but the worst thing is that it would make methods that return references (like `embeddings_ith`) be `unsafe`, since we wouldn't be able to ensure that the context isn't mutated while the returned reference is alive.

So I tend to think that using `&mut LlamaContext` is probably the desirable thing, even though it's a bit annoying.

madsmtm:
There might be a way to make the context getters use `&` again: add a `fn synchronize(&mut self) -> SynchronizedContext<'_>` helper, which then has the methods `fn embeddings_ith(&self, i: i32) -> &[f32]` to get the data out.

PR #809, opened 2025-08-23
feat: Add RPC backend support for distributed inference

hongkongkiwi:
## Summary
- Adds complete RPC backend support for distributed inference across multiple machines
- Includes `RpcBackend`, `RpcDevice`, and `RpcServer` classes for client/server operations
- Adds `LlamaSplitMode` enum for multi-GPU tensor parallelism configuration
- Includes comprehensive example demonstrating RPC client/server usage

## Features Added
- **RPC Backend**: Connect to remote llama.cpp RPC servers for distributed inference
- **Device Management**: Add/query remote RPC devices with memory information
- **Server Support**: Start RPC servers programmatically with configurable parameters
- **Split Modes**: Configure tensor parallelism across multiple GPUs (None, Layer, Row)
- **Complete Example**: Working client/server example with full CLI interface

## API Changes
- New `rpc` module with `RpcBackend`, `RpcDevice`, `RpcServer` structs
- `LlamaSplitMode` enum (None, Layer, Row) for multi-GPU configurations
- `LlamaModelParams::with_split_mode(split_mode: LlamaSplitMode) -> Self`
- `rpc` feature flag to enable RPC functionality

## Test Plan
- [x] Builds successfully with `--features rpc` flag
- [x] API matches upstream ggml RPC functions
- [x] Example compiles and runs without warnings
- [x] All documentation passes linting checks
- [x] Proper error handling and memory management

## Technical Details
The implementation uses the underlying GGML RPC functions (`ggml_backend_rpc_init`, `ggml_backend_rpc_add_device`, `ggml_backend_rpc_start_server`) from llama.cpp, providing safe Rust wrappers with comprehensive error handling.

RPC support enables:
- Distributed model execution across multiple machines
- Remote GPU/CPU resource utilization  
- Network-based tensor operations
- Multi-GPU tensor parallelism configurations

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1129, opened 2026-09-07
feat(params): `with_tensor_split` to set each device's share of the model

hxphsts:
## What

Adds `LlamaModelParams::with_tensor_split(&[f32])`, the safe setter for `llama_model_params::tensor_split`: the proportion of the model each device in `with_devices` receives. The slice is copied into the params struct's fixed-size array (`LLAMA_MAX_DEVICES`), with the remaining entries zeroed, so the caller never touches raw pointers or the array length. Also exposes the current value via `tensor_split()`.

## Why

`with_devices` selects which backends load a model, but without `tensor_split` llama.cpp decides the proportions by free memory alone. A caller placing a model deliberately (a small GPU beside a large one, or a device that should hold a fixed share) has no way to say so from the Rust side today, although `llama-cli --tensor-split` exposes it. This is the last piece of `llama_model_params` around device placement not reachable from the safe API.

Scope is deliberately just the params setter, in the spirit of the "safe API to the core llama.cpp features" scoping in #809.

## Test

`cargo check -p llama-cpp-2` on this branch; used in a downstream crate that splits a 26B model across two machines' GPUs by explicit proportions.

### DIFF
--- a/llama-cpp-2/src/model/params.rs
+++ b/llama-cpp-2/src/model/params.rs
@@ -559,6 +559,44 @@ impl LlamaModelParams {
         self
     }
 
+    /// Sets each device's share of the model, parallel to
+    /// [`with_devices`](Self::with_devices).
+    ///
+    /// Shares are relative, so they need not sum to one. An empty slice clears the split and
+    /// restores llama.cpp's default, which distributes in proportion to free memory. The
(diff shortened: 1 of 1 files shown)


MarcusDunn:
needs a fmt

PR #1143, opened 2026-09-19
fix(build): do not require CARGO_CFG_TARGET_FEATURE

pkubaj:
cargo exports CARGO_CFG_TARGET_FEATURE only when the target has at least
one enabled target feature. That is always true on x86_64 and aarch64,
but not on powerpc64/powerpc64le, where no target feature is stable, so
the variable is absent and the build script panics:

    Env var CARGO_CFG_TARGET_FEATURE not found.: NotPresent

The value is only used to map x86 features to GGML_* switches, so an
empty list is the correct result when the variable is missing.

Found while building on FreeBSD/powerpc64le and FreeBSD/powerpc64, where
llama.cpp itself builds and links fine once the script gets past this.


madsmtm (approved):
👍

AsbjornOlling (approved):

